PR #44, opened 2015-01-05
Update SpscOffHeapIntQueue.java

phaynes:
peekInt is modifying the queue incorrectly?


nitsanw:
Can you please explain why you think this is incorrect?
The consumer is updating the cached view of the producer index, there's nothing wrong with that.

phaynes:
My assumption is peek operation should be idempotent.
the setTailCache(getTail()) operation changes the queue state.

by deleting this line and changing

if (currentHead >= getTailCache()) {

With 
              if (currentHead >= getTail()) {

Idempotent operation is maintained.

nitsanw:
peek is idempotent in the sense that the consumer index is not modified and the buffer is not modified. That's the interface. There's no point to my mind in repeatedly failing the same condition over the state of the cached field. So peekInt is modifying the queue internal state, not invalidating the external API, and is to my mind correct.

PR #46, opened 2015-01-13
Draft of Blocking Queue

georges-gomes:
Hi Nitsan,

This is not for merge - just to open the discussion about efficient BlockingQueue implementations.

The SpscArrayQueueBlocking here is hand-crafted by inheritance of SpscArrayQueue but ultimately I would prefer to have the BlockingQueue Implementations dynamically generated on demand by the QueueFactory.

Let me know your thoughts before I push it too far.

Cheers
Georges

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/takestrategy/ParkTakeStrategy.java
+++ b/jctools-core/src/main/java/org/jctools/queues/takestrategy/ParkTakeStrategy.java
@@ -7,7 +7,6 @@ public class ParkTakeStrategy<E> implements TakeStrategy<E>
 {
 
     public volatile int storeFence = 0;
-    public volatile int loadFence = 1;
 
     private AtomicReference<Thread> t               = new AtomicReference<Thread>(null);
 
@@ -15,9 +14,9 @@ public class ParkTakeStrategy<E> implements TakeStrategy<E>
     public void signal()
     {
         // Make sure the offer is visible before unpark
-        storeFence = loadFence; // JVM8 -> UNSAFE.storeFence();
+        storeFence = 1; // store barrier
 
-        LockSupport.unpark(t.get());
+        LockSupport.unpark(t.get()); // t.get() load barrier
     }
 
     @Override
--- a/jctools-core/src/main/java/org/jctools/queues/takestrategy/ParkTakeStrategy.java
+++ b/jctools-core/src/main/java/org/jctools/queues/takestrategy/ParkTakeStrategy.java
@@ -1,0 +1,43 @@
+package org.jctools.queues.takestrategy;
+
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.LockSupport;
+
+public class ParkTakeStrategy<E> implements TakeStrategy<E>
+{
+
+    public volatile int storeFence = 0;
+    public volatile int loadFence = 1;
+
+    private AtomicReference<Thread> t               = new AtomicReference<Thread>(null);
+
+    @Override
+    public void signal()
+    {
+        // Make sure the offer is visible before unpark
+        storeFence = loadFence; // JVM8 -> UNSAFE.storeFence();
+
+        LockSupport.unpark(t.get());
+    }
+
+    @Override
+    public E waitFor(SupplierJDK6<E> supplier)
+    {
+        E e = supplier.get();
+        if (e != null)
+        {
+            return e;
+        }
+
+        t.set(Thread.currentThread());
+
+        while ((e = supplier.get()) == null)
+        {
+            LockSupport.park();
+        }
+
+        t.lazySet(null);
+
+        return e;
+    }
+}
--- a/jctools-core/src/main/java/org/jctools/queues/takestrategy/ParkTakeStrategy.java
+++ b/jctools-core/src/main/java/org/jctools/queues/takestrategy/ParkTakeStrategy.java
@@ -3,7 +3,7 @@ package org.jctools.queues.takestrategy;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.LockSupport;
 
-public class ParkTakeStrategy<E> implements TakeStrategy<E>
+public final class ParkTakeStrategy<E> implements TakeStrategy<E>
 {
 
     public volatile int storeFence = 0;
--- a/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
+++ b/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
@@ -1,5 +1,7 @@
 package org.jctools.queues.takestrategy;
 
+import org.jctools.queues.spec.ConcurrentQueueSpec;
+
 import java.util.Queue;
 
 public final class YieldTakeStrategy<E> implements TakeStrategy<E>
@@ -19,5 +21,11 @@ public final class YieldTakeStrategy<E> implements TakeStrategy<E>
             Thread.yield();
         }
         return e;
+    }
+
+    @Override
+    public boolean supportsSpec(ConcurrentQueueSpec qs)
+    {
+        return true;
     }
 }
--- a/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
+++ b/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
@@ -1,5 +1,7 @@
 package org.jctools.queues.takestrategy;
 
+import java.util.Queue;
+
 public final class YieldTakeStrategy<E> implements TakeStrategy<E>
 {
     @Override
@@ -9,10 +11,10 @@ public final class YieldTakeStrategy<E> implements TakeStrategy<E>
     }
 
     @Override
-    public E waitFor(SupplierJDK6<E> supplier) throws InterruptedException
+    public E waitFor(Queue<E> q) throws InterruptedException
     {
         E e;
-        while((e = supplier.get()) == null)
+        while((e = q.poll()) == null)
         {
             Thread.yield();
         }
--- a/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
+++ b/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
@@ -1,6 +1,6 @@
 package org.jctools.queues.takestrategy;
 
-public class YieldTakeStrategy<E> implements TakeStrategy<E>
+public final class YieldTakeStrategy<E> implements TakeStrategy<E>
 {
     @Override
     public void signal()
@@ -11,15 +11,11 @@ public class YieldTakeStrategy<E> implements TakeStrategy<E>
     @Override
     public E waitFor(SupplierJDK6<E> supplier)
     {
-        while(true) // Should introduce safepoints
+        E e;
+        while((e = supplier.get()) == null)
         {
-            E e = supplier.get();
-            if (e!=null)
-            {
-                return e;
-            }
-
             Thread.yield();
         }
+        return e;
     }
 }
--- a/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
+++ b/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
@@ -1,0 +1,25 @@
+package org.jctools.queues.takestrategy;
+
+public class YieldTakeStrategy<E> implements TakeStrategy<E>
+{
+    @Override
+    public void signal()
+    {
+        // Nothing to do
+    }
+
+    @Override
+    public E waitFor(SupplierJDK6<E> supplier)
+    {
+        while(true) // Should introduce safepoints
+        {
+            E e = supplier.get();
+            if (e!=null)
+            {
+                return e;
+            }
+
+            Thread.yield();
+        }
+    }
+}
--- a/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
+++ b/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
@@ -11,7 +11,7 @@ public final class YieldTakeStrategy<E> implements TakeStrategy<E>
     }
 
     @Override
-    public E waitFor(Queue<E> q) throws InterruptedException
+    public E waitPoll(Queue<E> q) throws InterruptedException
     {
         E e;
         while((e = q.poll()) == null)
--- a/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
+++ b/jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
@@ -9,7 +9,7 @@ public final class YieldTakeStrategy<E> implements TakeStrategy<E>
     }
 
     @Override
-    public E waitFor(SupplierJDK6<E> supplier)
+    public E waitFor(SupplierJDK6<E> supplier) throws InterruptedException
     {
         E e;
         while((e = supplier.get()) == null)
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ jctools-parent.iml
 
 
 /.settings
+
+target/MANIFEST.MF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,0 +1,10 @@
+First release!
+==========
+JCTools offers java.util.Queue implementations for concurrent use. The queues
+support a limited subset of the interface focusing on message passing rather
+than collection like semantics. In particutlar the queues do not support:
+- The iterator() method
+- remove()
+- contains()
+
+Enjoy!
--- a/jctools-benchmarks/pom.xml
+++ b/jctools-benchmarks/pom.xml
@@ -24,12 +24,12 @@
 		<dependency>
 			<groupId>org.openjdk.jmh</groupId>
 			<artifactId>jmh-core</artifactId>
-			<version>1.0</version>
+			<version>1.4</version>
 		</dependency>
 		<dependency>
 			<groupId>org.openjdk.jmh</groupId>
 			<artifactId>jmh-generator-annprocess</artifactId>
-			<version>1.0</version>
+			<version>1.4</version>
 			<scope>provided</scope>
 		</dependency>
 	</dependencies>
--- a/jctools-benchmarks/pom.xml
+++ b/jctools-benchmarks/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.jctools</groupId>
 		<artifactId>jctools-parent</artifactId>
-		<version>1.0-SNAPSHOT</version>
+		<version>1.0</version>
 	</parent>
 	<artifactId>jctools-benchmarks</artifactId>
 	<packaging>jar</packaging>
--- a/jctools-benchmarks/src/main/java/org/jctools/handrolled/throughput/spsc/BusyQueuePerfTest.java
+++ b/jctools-benchmarks/src/main/java/org/jctools/handrolled/throughput/spsc/BusyQueuePerfTest.java
@@ -13,6 +13,7 @@
  */
 package org.jctools.handrolled.throughput.spsc;
 
+import org.jctools.handrolled.throughput.spsc.QueuePerfTest.Producer;
 import org.jctools.queues.QueueByTypeFactory;
 
 import java.util.Queue;
@@ -39,48 +40,52 @@ public class BusyQueuePerfTest {
         System.out.format("summary,BusyQueuePerfTest,%s,%d\n", queue.getClass().getSimpleName(), sum / 10);
     }
 
-    private static long performanceRun(final int runNumber, final Queue<Integer> queue) throws Exception {
-        final long start = System.nanoTime();
-        final Producer p = new Producer(queue);
-        final Thread thread = new Thread(p);
-        thread.start();
+    private static long performanceRun(int runNumber, Queue<Integer> queue) throws Exception {
+        Producer p = new Producer(queue);
+        Thread thread = new Thread(p);
+        thread.start();// producer will timestamp start
 
         Integer result;
         int i = REPETITIONS;
-        int f = 0;
+        int queueEmpty = 0;
         do {
             while (null == (result = queue.poll())) {
-                f++;
+                queueEmpty++;
             }
         } while (0 != --i);
-
+        long end = System.nanoTime();
+        
         thread.join();
-
-        final long duration = System.nanoTime() - start;
-        final long ops = (REPETITIONS * 1000L * 1000L * 1000L) / duration;
-        System.out.format("%d - ops/sec=%,d - %s result=%d failed.poll=%d failed.offer=%d\n",
-                Integer.valueOf(runNumber), Long.valueOf(ops),
-                queue.getClass().getSimpleName(), result,f,p.fails);
+        long duration = end - p.start;
+        long ops = (REPETITIONS * 1000L * 1000L * 1000L) / duration;
+        String qName = queue.getClass().getSimpleName();
+        System.out.format("%d - ops/sec=%,d - %s result=%d failed.poll=%d failed.offer=%d\n", runNumber, ops,
+                qName, result, queueEmpty, p.queueFull);
         return ops;
     }
 
     public static class Producer implements Runnable {
         private final Queue<Integer> queue;
-        int fails=0;
-        public Producer(final Queue<Integer> queue) {
+        int queueFull = 0;
+        long start = 0;
+
+        public Producer(Queue<Integer> queue) {
             this.queue = queue;
         }
 
         public void run() {
-            final Queue<Integer> q = queue;
             int i = REPETITIONS;
-            int f=0;
+            int f = 0;
+            Queue<Integer> q = queue;
+            long s = System.nanoTime();
             do {
                 while (!q.offer(TEST_VALUE)) {
                     f++;
                 }
             } while (0 != --i);
-            fails = f;
+            
+            queueFull = f;
+            start = s;
         }
     }
 }
--- a/jctools-benchmarks/src/main/java/org/jctools/jmh/throughput/QueueThroughputBackoffNone.java
+++ b/jctools-benchmarks/src/main/java/org/jctools/jmh/throughput/QueueThroughputBackoffNone.java
@@ -29,6 +29,7 @@ public class QueueThroughputBackoffNone {
     private static final long DELAY_PRODUCER = Long.getLong("delay.p", 0L);
     private static final long DELAY_CONSUMER = Long.getLong("delay.c", 0L);
     private static final Integer ONE = 777;
+    private Integer escape;
     @Param(value = { "SpscArrayQueue", "MpscArrayQueue", "SpmcArrayQueue", "MpmcArrayQueue" })
     String qType;
     @Param(value = { "132000" })
@@ -93,19 +94,21 @@ public class QueueThroughputBackoffNone {
 
     @Benchmark
     @Group("tpt")
-    public Integer poll(PollCounters counters, ConsumerMarker cm) {
+    public void poll(PollCounters counters, ConsumerMarker cm) {
         Integer e = q.poll();
         if (e == null) {
             counters.pollsFailed++;
             backoff();
         }
-        else {
+        else if (e == ONE){
             counters.pollsMade++;
+        }
+        else {
+            escape = e;
         }
         if (DELAY_CONSUMER != 0) {
             Blackhole.consumeCPU(DELAY_CONSUMER);
         }
-        return e;
     }
 
     @TearDown(Level.Iteration)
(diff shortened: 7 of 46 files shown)


ChristianWulf:
Good job so far. Here are a few hints:
* The following code should be instantiated only once at construction and cached in a final field:
```java
new SupplierJDK6<E>() {
@Override
public E get() {
return poll();
}
}
```
* All take strategies should be declared final

nitsanw:
I'm with you on codegen. What we want is a class Q to be extended by a generic blocking stategy to give us some sort of mixin of strategy and lock free queue. We can try and do that with @RichardWarburton templates.
I think we should start simpler with a generic wrapper which takes a strategy and queue, delegates all the queue methods and adds the blocking queue methods. Once we've worked out correctness and blocking strategies we can use the templates to generate the queues.

jctools-core/src/main/java/org/jctools/queues/takestrategy/YieldTakeStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public void signal()
+    {
+        // Nothing to do
+    }
+
+    @Override
+    public E waitFor(SupplierJDK6<E> supplier)
+    {
+        while(true) // Should introduce safepoints
```
nitsanw: Thread.yield() should introduce the safepoint, no?
In any case this loop will be just as correct written naturally:

    E e;
    while((e = s.get()) == null)
      Thread.yield();
    return e;

jctools-core/src/main/java/org/jctools/queues/takestrategy/ParkTakeStrategy.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public volatile int storeFence = 0;
+
+    private AtomicReference<Thread> t               = new AtomicReference<Thread>(null);
+
+    @Override
+    public void signal()
+    {
+        // Make sure the offer is visible before unpark
+        storeFence = 1; // store barrier
```
nitsanw: Why is this required? shouldn't the queue semantics provide for correct offer visibility?

georges-gomes: In the case of SPSC, the element is offered in the array with "putOrdered".
It's important that the unpark "happen after".
Otherwise there is a risk for the waiting thread to be unpark without seeing the element.
And so being reparked without processing the element.

Actually, you raised that problem to me a few months ago :)
I fixed it with a "putVolatile" in my "SPSCBlockingQueue" and I'm using a fence here to untouch the performance of the SPSC Queue when used in a non-blocking way.

nitsanw: I see, so you are not actually looking for a store fence, rather for a StoreLoad to prevent the load of the parked thread moving up.
I was trying to remember what the original problem looked like, but a bit brain dead this week.

nitsanw:
Thinking out loud, we have 2 threads, here's a simplified view of what's happening:
P: {
  offer(e);// this is an effective StoreStore
  dummyStoreLoad;// the store to the queue can't get re-ordered with the load of parkedThread
  unpark(parkedThread.get());
}
C: {
  if(e = poll() != null) return e;// This is an optimization, in terms of correctness we can skip
  parkedThread.set(Thread.current()); //StoreLoad, the loads after parkedThread is set can't move up
  while (e = poll() == null)
    park();
  parkedThread.set(null);
}

So the question:
"Can a put to the queue fail to unpark the consumer?"
Can be simplified to the following(v,u are volatile fields):
P {v = 1; if (u == 1) unpark(C);}
C {u = 1; if (v == 0) park(C); u = 0;}

I think it should work, but having a bad week in terms of sleep, so not trusting myself much.

georges-gomes:
That's exactly it.

georges-gomes:
@nitsanw could you point me to the entrypoints of the templates of @RichardWarburton?
Thanks

nitsanw:
Have a look at this commit from @RichardWarburton: https://github.com/JCTools/JCTools/commit/d8d9a4c07fcccce1c8be846d67237851d6bfa777

RichardWarburton:
You can also see the class in question at: https://github.com/JCTools/JCTools/blob/master/jctools-experimental/src/main/java/org/jctools/util/Template.java

And simple examples of usage at https://github.com/JCTools/JCTools/blob/master/jctools-experimental/src/test/java/org/jctools/util/TemplateTest.java

georges-gomes:
The Template and SimpleCompiler classes are pretty cool. Thanks @RichardWarburton!
This is what I made today in a few hours today.
Blocking queues on demand.

Next step: Mixin Take and Put strategies?

Your feedback welcomed.

Cheers
GG

nitsanw:
Thanks for all the hard work, this is looking pretty good.
Have you run any benchmarks to compare with ArrayBlockingQueue?

georges-gomes:
I have done some benchmark internally and this form is lighter, probably just by the way it signal.
The latency is also better between producer and consumer wakeup.
I have a few JMH benches that I will port for this JCTools implementation... 
Always better to see these things live and sometimes on more than one machine :)

ChristianWulf:
I do not see the benefits of the dynamic creation. Why don't you generate blocking implementations as java files? Moreover, a blocking implementation is currently always generated, even if it has been generated once or more before.

nitsanw:
@ChristianWulf dynamic creation can help in minimising the size of the deployed jar especially as more parameters are added to the generated flavours. It can also offer some interesting optimisation options that are otherwise not easy to achieve (e.g. several padding configurations for same queue).
I agree that the generated classes can and should be cached to avoid bloat and allow compiled code to be reused.

RichardWarburton:
Caching should be evaluated on a case-by-case basis. Specifically code generating new instances can give the JIT compiler opportunities for optimisation that wouldn't be possible otherwise. Obviously it also has a potential tradeoff in terms of additional memory consumption and JIT overhead. So the ability to generate new instances vs cached instances is something that end users looking for optimal performance might want to have control over.

georges-gomes:
Should I just add a cache layer in from of it?

georges-gomes:
Cache done. Looks good to me ;)

ChristianWulf:
Why don't you cache the class, but the class cloader?

georges-gomes:
Just a small mistake. I will change it.

georges-gomes:
SpscArrayQueue
Benchmark                                Mode  Samples  Score   Error  Units
o.j.j.b.SingleThreadedOffer.offerLoop    avgt        5  2.008 ± 0.039  ns/op
o.j.j.b.SingleThreadedPoll.poll          avgt        5  1.934 ± 0.065  ns/op

ArrayBlockingQueue
Benchmark                                Mode  Samples   Score   Error  Units
o.j.j.b.SingleThreadedOffer.offerLoop    avgt        5  17.410 ± 1.527  ns/op
o.j.j.b.SingleThreadedPoll.poll          avgt        5  17.155 ± 1.421  ns/op

BLOCKING SPSC
Benchmark                                Mode  Samples  Score   Error  Units
o.j.j.b.SingleThreadedOffer.offerLoop    avgt        5  9.003 ± 1.006  ns/op
o.j.j.b.SingleThreadedPoll.poll          avgt        5  1.942 ± 0.131  ns/op

BLOCKING MPMC
Benchmark                                Mode  Samples   Score   Error  Units
o.j.j.b.SingleThreadedOffer.offerLoop    avgt        5  15.235 ± 1.970  ns/op
o.j.j.b.SingleThreadedPoll.poll          avgt        5  14.316 ± 2.765  ns/op

georges-gomes:
We can see the cost of the store fence and the signal in the offer

ChristianWulf:
Your first results look reasonable. Good job!

But now the non-blocking queue variants are not final anymore leading to a potential performance  decrease. Why can't we avoid inheritance and use a wrapper class instead to instantiate a blocking variant?

For example:
```java
public final class PCBlockingQueue<E> implements BlockingQueue<E> {

  private final Queue<E> q;
  private final TakeStrategy takeStrategy;

  public PCBlockingQueue(Queue<E> q, Class<TakeStrategy> takeStrategyClass) {
    this.q = q;
    // avoid reflection in production mode; example should just show the wrapper alternativ
    this.takeStrategy = takeStrategyClass.newInstance(q);  // using q.poll internally
  }

 @Override
public void put(E e) throws InterruptedException
{
while(!q.offer(e))
{
Thread.yield();
}
}

@Override
public E take() throws InterruptedException
{
return takeStrategy.waitFor(poller); // For JVM8 => return takeStrategy.waitFor(this::poll);
}

...

}
```

Usage:
```java
Queue<X> q = QueueFactory.newQueue(..);
BlockingQueue<X> bq = new PCBlockingQueue<X>(q, ParkTakeStrategy.class);
```

georges-gomes:
I don't think that there is any optimisation with final classes on existing VMs.
Call sites are optimized and deoptimized based on the instance types encountered.

Anyone else thinks like @ChristianWulf ?

RichardWarburton:
Its a little more complex than that. In the case of a monomorphic callsite
of a nonfinal class the jit compiler needs to insert a check on the type of
the receiver object and jumpt to a deoptimisation path if it fails. See
http://insightfullogic.com/blog/2014/may/12/fast-and-megamorphic-what-influences-method-invoca/
for details and some simple benchmarking.

So final saves you a predictable branch per callsite. A little but not much
cost.

georges-gomes:
Regarding the BlockingQueue wrapping the Queue: I tried already. The performance is impact, quite significantly by the referencing. Even final, again JIT doesn't produce the aggressiveness expected.
I even want to inline the "TakeStrategy" code through the template instead of calling it.
I should be able to demonstrate that in a minute...

georges-gomes:
Inline the ParkTakeStrategy code manually

Benchmark                                Mode  Samples  Score   Error  Units
o.j.j.b.SingleThreadedOffer.offerLoop    avgt        5  8.422 ± 0.171  ns/op
o.j.j.b.SingleThreadedPoll.poll          avgt        5  1.956 ± 0.119  ns/op

Maybe not visible in real life...

ChristianWulf:
I proposed the final keyword for the class to implicitly declare all methods as final to encourage method inlining. However, I have not found any serious reference confirming that final methods are inlined. Can anyone give a reference?

RichardWarburton:
Methods aren't inlined just because they are final.

Methods are inlined according to inlining being safe and some heuristic. Methods being final is a sufficient property for inlining to be safe, but so is a callsite being monomorphic. Heuristics include:

 * Size of the method (35 bytes for regular methods, 325 if they are considered "very hot")
 * Number of method calls which have been inlined already (ie the depth of the calltree which has already been called). This has to be < 9.

These are hotspot specific numbers, all configurable at runtime. Other JVMs may perform completely differently.

The blog post which I linked earlier explains the different in performance you'll see in situations where methods are and aren't inlined.

ChristianWulf:
Yes, I read your blog post. And yes, I know that a final method is not always inlined. That's why I wrote "encourage". But where did you find the following statement?
"Methods being final is a sufficient property for inlining to be safe, ..."

RichardWarburton:
You're right to pick me on that particular point - because its not true! What I should have said was that a method being final was part of the safety criteria, but it didn't need to be final to be inlined - the callsite being monomorphic would do just as well. For example recursive methods need to be dealt with differently.

As to sources of information in general a combination of reading source code, books, talking to people and experimentation. There is no up to date, accurate and canonical source of information about the topic. We should all ask @nitsanw to blog about it and provide one though ;)

nitsanw:
@RichardWarburton @ChristianWulf @georges-gomes This discussion is going so well!!! thanks for spending the time guys, I'm swamped with work/family at the moment and will devot some time to this either over the weekend or sometime next week.
As for a blog post on inlining... I think Mr. JITWatch (AKA @chriswhocodes) could write a good one

ChristianWulf:
I agree. The discussion is great! However, we have not yet finished it. I think, we need further benchmarks to compare the inheritance version with the wrapper version.
In particurlar, I want to ensure that the used benchmarks

- include a warm-up phase,
- take a sufficient amount of time (to reach a steady-state),
- discard the first half measurements (to further reduce the effect of unoptimized compiled code), and
- have a small 95%-confidence interval.

Everything needs to be true to represent a valid performance evaluation. Can anyone explain the methodology of the used benchmarks? Do they fulfil the requirements?

nitsanw:
Benchmarks are JMH(look in jctools-benchmarks) so warmup etc is easily tunable.

Discarding results is not a practice I side with, pick a warmup period.

JMH reports err to help inform you on that. It's important to run multiple forks to catch memory layout impacts. Watch out for multi-modal behaviour getting averaged out in the end result.

georges-gomes:
Just committed some improvements on testing and some bug fixes.

@ChristianWulf 's wrapping proposal is likely to be slower for blocking queue (very very slightly).
But it keep the "final" modifier on non-blocking queues.

Where you guys want to take it?
I really don't care, most of the cost is in the waiting / signal code anyway.

Cheers
GG

jpbempel:
@RichardWarburton @ChristianWulf 
Yes Richard is right it is little bit more complex than that.
In fact if you are calling a method trough an interface, whether the class implemented the interface is final or not just not matter, and the guard on type is inserted anyway.
if you are calling the method from a variable with same type of the instance, if there is no guard inserted as long as there is no subclasses loaded by the JVM.
It is possible because of the CHA and if a subclasses is loaded deopt is triggered on the callsite involved with this ancestors.
So for me no need to put final on the class.

Also take a look at my blog post on this subject: http://jpbempel.blogspot.com/2012/10/virtual-call-911.html

Cheers
Jean-Philippe

nitsanw:
On that note, all the JCTools benchmarks use the queues via an interface and test one queue at a time, so the call site always has the check

RichardWarburton:
Is that the wrong behaviour though? If end users are expected to call the methods via the interface then that's what should be benchmarked and if that's the expected API interaction there's no point in worrying about making things final from a performance point of view.

nitsanw:
It is neither right nor wrong, it just is. <mystic smile>
The benchmarks reflect a common usecase where JCTools queues are drop in replacements for JDK queues. They also reflect a certain amount of justified laziness, I'm reluctant to C&P a benchmark per queue, and it is nice that JMH can collect the stats for all types in one report for me.
The QueueFactory reflects that same usage pattern as well.
The end-users are free to do what they will and I expect some of them will be using direct classes.
My concern with making the classes final is that it forces users to compose rather than inherit which is sadly a data layout decision as well as an OO decision.
It's an interesting discussion and it's worth quantifying the delta for this use case.

ChristianWulf:
Ok. It is not that easy with the `final` keyword. I see.
However, the wrapper implementation also reduces the code size resulting in a more clearly code structure. I just throw in this argument so that we can discuss it.

georges-gomes:
+1 it's a major advantage.

ChristianWulf:
Another question concerning the *morphic callsites: Consider a huge Java application where several libraries use at least 3 implementations per interface/abstract class. Then, almost all method calls are compiled to megamorphic calls, right?

ChristianWulf:
@jpbempel 
You wrote the following:
"Also getter and setter of object will be largely optimized by monomorphic call followed by inlining !"

This can thus only be true if the variable's class is not declared as an interface and or super class, right?

And due to another statement of you, inlining is excluded for almost all programs (at least for all my programs) that use interfaces and super classes for their variable declarations since corresponding method calls would be compiled to virtual calls, right?
"Inlining cannot be perform with regular virtual calls but only on classic calls."

chriswhocodes:
Hi @ChristianWulf inlining is an important early step for the JIT compiler as it "expands the horizon" for further optimisations by bringing related code closer together (e.g. before loop unrolls).

Methods called on an interface can certainly be inlined if they are observed to be monomorphic or bimorphic at runtime (you can play around with this in the JITWatch sandbox example https://github.com/AdoptOpenJDK/jitwatch/blob/master/src/main/resources/examples/PolymorphismTest.java).

If a method cannot be inlined (it fails the JIT compiler inlining rules due to size/depth etc) the virtual call can still be optimised (e.g. invokeinterface to invokevirtual) if it's mono or bimorphic. (Look for `optimized virtual_call` in assembly comments on a callq).

Once things get megamorphic then I believe these optimisation are off the table.

John Rose did a good JIT write-up here: https://wikis.oracle.com/display/HotSpotInternals/PerformanceTechniques and Martin Thompson did some benchmarking on method dispatch here: http://mechanical-sympathy.blogspot.co.uk/2012/04/invoke-interface-optimisations.html

Hopefully between all of this and @RichardWarburton and @jpbempel 's blogs we can find the truth!

leventov:
@RichardWarburton

> Caching should be evaluated on a case-by-case basis. Specifically code generating new instances can give the JIT compiler opportunities for optimisation that wouldn't be possible otherwise.

What optimizations are possible when loading a class each time instead of caching?

ChristianWulf:
Hi @chriswhocodes,
so the statement "Inlining cannot be perform with regular virtual calls but only on classic calls." is wrong, right? Or what is meant by "regular virtual call"?

ChristianWulf:
@chriswhocodes 
The PolymorphismTest is very interesting. I conclude that a method call is detected as mono- or bimorphic depending on the current context and statistics of the JVM. It does matter whether the JVM has loaded more than two implementations of an interface or not. It depends on how many implementations are currently used within the current context (e.g. within a loop), right?

For example, the `PolymorphismTest` instantiates three implementations of the `Animal` interface, but uses only one or two implementations within the loop if we set `maxImplementations` to 1 or, respectively, two.

jpbempel:
@ChristianWulf 
Devirtualization (calling directly a method instead of resolving the virtual call) can happen at each call site so wether this is a variable of the class or an interface you can have the optimzed virtual call in both case.
If you have a method that take an interface List, but only an ArrayList instance is always pass to this method, the call will optimized as monomorphic to the corresponding ArrayList method.

Inlining cannot be performed through virtual call.
But if the call can be devirtualized then it is not a virtual call anymore but a classic one so inlining can be performed and as @chriswhocodes mention, this is very important to perform other optimizations.

chriswhocodes:
@ChristianWulf exactly, HotSpot determines whether a method call can be devirtualized (and therefore inlined) by observing at runtime. You can declare as many variables as you like of the interface type but the decision to devirtualize is made on how many different concrete implementations were seen.

This is a speculative optimisation based on runtime observations and the JIT compiler inserts an uncommon trap so that it can de-optimise if the speculation turned out to be wrong (e.g. more than 2 implementations of the interface were encountered).

ChristianWulf:
@jpbempel and @chriswhocodes 
Nice summary! So methods can only be inlined when invoked on an variable of the concrete class (i.e., no virtual call) or when detected as mono- or bimorphic call (i.e., an optimized virtual call).

ChristianWulf:
So, which variant of the blocking queue implementations should we use? And could you, Nitsan, integrate the resulting variant into JCTools soon?

ChristianWulf:
If someone likes to know what I need and use JCTools in general and a blocking queue in particular for, have a look at http://teetime.sourceforge.net. I use the queues as basis for the pipes in our pipes-and-filters framework.

georges-gomes:
What do you guys think if I do the simple wrapping implementation in the
jctools-core and move the template generation version into
jctools-experimental? Then, we look at again later.

ChristianWulf:
+1 for @georges-gomes's proposal

nitsanw:
Note: Unbounded queues CAN be blocking. In particular the producer will never block, but the consumer will if the queue is empty.

georges-gomes:
Absolutely, I will fix that.

nitsanw:
Working through the code, making notes as I go:

1. Blocking refers to both take and put, I notice in the template we only support a configurable policy on the take side. put() defaults to a: "while(!offer(e)) Thread.yield()". We need to allow configurable wait strategies on either side.
2. Given we're code generating on the fly we can have a JDK8 template that avoids the function interface required in JDK7/6.
3. Given the only utility of the Supplier interface is to call 'poll()' we can do away with the interface and pass in the Queue, i.e. 'this'.
4. The drain methods are important (at least from personal experience they seem to be in common use). They are also easy to implement: "while((e = poll()) != null) c.add(e);". Is there a good reason to avoid?
5. We use a volatile store dummy variable and an atomic reference in the SCParkTake. I think we can improve by using getAndSet:

------

     private final AtomicReference<Thread> t = new AtomicReference<Thread>(null);
     @Override
     public void signal()
     {
        Thread cThread = t.getAndSet(null); // store load barrier
		LockSupport.unpark(cThread); 
     }
     @Override
     public E waitFor(Queue<E> q)
     {
        E e;
        if ((e = q.poll()) != null)
        	return e;
        t.set(Thread.currentThread());
        // can spin on t.get() != null?
        while ((e = q.poll()) == null)
        	LockSupport.park();
        return e;
     }

ChristianWulf:
Some refactoring:
```java
 @Override
 public E waitFor(Queue<E> q)
 {
    E e;
    if ((e = q.poll()) == null) {
      t.set(Thread.currentThread());
      // can spin on t.get() != null?
      while ((e = q.poll()) == null)
        LockSupport.park();
    }
    return e;
 }
```
If this code is too large for inlining, we could refactor the true-case into another method:
```java
 @Override
 public E waitFor(Queue<E> q)
 {
    E e;
    if ((e = q.poll()) == null) {
      e = park(q);
    }
    return e;
 }
```

jpbempel:
@ChristianWulf 
Several things regarding your comment about inlining:

35 bytes method size is not the only parameter to trigger inlining of a method.
If the method is HOT (executed a lot) the method can also be inlined if size over 35 bytes but below 350 bytes (by default).
If it is true that using refactorization into separate method can reduce the byte code size of a method, it helps significantly if the branch is taken only in exceptional cases otherwise the refactored method will be inlined anyway on the caller method and we end up with the same thing.
I mean here it should not be the only reason to factorize like this, can be readability (although for such small code it is better to have it under the same eyes than the rest of the method).

Thanks.

georges-gomes:
getAndSet implementation works but it's slightly slower.
8.350 ns vs 8.200ns - JMH repeated many times
(for the uncontended / single-threaded case)

georges-gomes:
I still need to find a way to expose the selection of Put and Take strategy to the "user".

ChristianWulf:
A string is not type safe; an enum prevents the addition of further strategies. So, passing the strategy's class object is a solution.

daschl:
Great conversations going on, and since @nitsanw pointed me to this ticket I'd like to chime in with my use case - maybe you can see if this fits or not?  :)

I'm currently using the disruptor in a "MPSC" setup, but it only allows you to configure the producer side (so it's set to multi). I'm using it for two specific things:
- Providing fast feedback backpressure to the producers if they outpace the consumer
- Batching semantics. Since downstream I'm writing to IO in netty I only flush() when i get the "end of batch" signal from the disruptor in my consumer.

Backpressure can easily be covered anyways, but are the batching semantics that make sense in that approach? Also, I guess making the waiting strategies configurable is a must (?) since I care about latency, but also I can't spin wait because this runs in a customer environment alongside with other services.

Looking forward to help / test / whatever makes sense for you folks :)

nitsanw:
I'm not a fan of the batch end approach in the disruptor as it takes the following form:

          final long availableSequence = sequenceBarrier.waitFor(nextSequence);
          while (nextSequence <= availableSequence)
          {
              event = dataProvider.get(nextSequence);
              eventHandler.onEvent(event, nextSequence, nextSequence == availableSequence);
              nextSequence++;
          }
          sequence.set(availableSequence);

This leads to:
 * Consumer sequence is updated infrequently, which means producers can stall despite plenty of room in buffer
 * End of batch signal does not mean no further elements are available, just that no further elements were visible when you started this batch
 * The win in updating the consumer index infrequently is very small. This is not true from the producer POV if consumer sequence is queried on each offer, but if the value is cached (as it is in JCTools) than I don't really see the point.

I would suggest you pick a batch size you are happy with because flushing every X messages makes sense, and use the stronger guarantee of poll == null as a further trigger.

daschl:
@nitsanw makes sense.. I just wonder what the right batch size is, maybe it could also be latency driven (flush after N µs or when written before and poll == null)

nitsanw:
you are describing a mini-nagles algo:
 * batch end by time since first message in batch
 * batch end by size of batch

For sending over network in less latency sensitive envs. it makes sense to wait for more messages before sending (as does nagles) 
This is interesting, but not within the scope of JCTools at the moment.

daschl:
yes, thanks - let's not pollute this topic here with it :)

ChristianWulf:
What about a SleepTakeStrategy where the consumer goes to bed for a configurable amount of time if the queue is empty?

nitsanw:
Seems like a specialization of the Park one?

ChristianWulf:
When looking at the `` park`` method, then: yes, you're right! I didn't know.
```public native void park(boolean isAbsolute, long time);```

**Comment by nitsanw** [comment]
In particular it seems redundant for a notification based implementation to wake up periodically as the interface does not allow the thread to 'escape' from the take/put methods.

**Comment by georges-gomes** [comment]
Hi @daschl,

Due to the single consumer relaxed constraint, the waiting Park strategy in this pull request is very fast to wake up. Faster than a object.notify or condition.signal/signalAll that you will found in a ArrayBlockingQueue. I needed something fast to wake up and we have benchmark this wake up time carefully (bench not provided here - but it's a single sleep -> round trip time on the queue).

The backpressure is pretty single as you said, just look for offer() returning false.

For the batching:

Let's create a batching interface

```java
interface BatchingConsumer
{
void startBatch()
void newElement(E e)
void endBatch()
}
```

In the consuming thread you do something like this:

```java
public void run()
{
    while(true)
    {
         // wait for a first element
         E e = queue.take();
         
         batchConsumer.startBatch();
         batchConsumer.newElement(e);
         
         // Spin with some yield
         for(int i=0; i<100;i++)
         {
              E e = queue.poll()
              if (e != null)
                    batchConsumer.newElement();
              else
                    Thread.yield();
         }

         batchConsumer.endBatch();
    }
}
```

This will make sure you don't have more than 100 elements in one batch.
The Thread.yield() will create some time to wait for data, not crazy spinning and not calculating time offs with System.nanotime() all over the place.

But I don't like to wait for data. I would rather do:

```java
public void run()
{
    while(true)
    {
         // wait for a first element
         E e = queue.take();
         
         batchConsumer.startBatch();
         batchConsumer.newElement(e);
         
         // Batch
         for(int i=0; i<100;i++)
         {
              E e = queue.poll()
              if (e != null)
                    batchConsumer.newElement();
              else
                    break;
         }

         batchConsumer.endBatch();
    }
}
```

Taking full advantage of the thread parking efficiency.

daschl:
@georges-gomes very interesting, thanks for sharing. Now I just need to find time and rip out the disruptor and see what I can get over this :)

ChristianWulf:
I found a bug in the ``waitPoll`` method. Although it is declared to throw an ``InterruptedException``, it does not since ``park()`` does not throw this exception. Thus, I added the if-statement after parking.
```java
	@Override
	public E waitPoll(final Queue<E> q) throws InterruptedException
	{
		E e = q.poll();
		if (e != null)
		{
			return e;
		}

		t.set(Thread.currentThread());

		while ((e = q.poll()) == null) {
			LockSupport.park();
			if (Thread.currentThread().isInterrupted()) {
				throw new InterruptedException("Interrupted while waiting for the queue to become non-empty.");
			}
		}

		t.lazySet(null);

		return e;
	}
```

georges-gomes:
Yes, you are right! I will add that shortly.

nitsanw:
Second review in progress, will comment as I go...
1. PutStrategy.backoffOffer - should throw interrupted, as dictated by the BlockingQueue API, and implementation should check for Thread.interrupted and throw for it.
2. instead of put/take strategy packages, maybe put both under one package called blocking?

nitsanw:
Just a note on the MCParkTakeStrategy that condition.await generates garbage. I hit a case yesterday where replacing LinkedBQ with ArrayBQ was generating more garbage because of that...

nitsanw:
Semi relevant discussion on park/unpark HB rules:
http://jsr166-concurrency.10961.n7.nabble.com/unpark-park-memory-visibility-td11812.html
I don't think it helps us any, because the problem is that we want to prevent the 't' field read from re-ordering with the offer:

   offer(e);
   // need a StoreLoad barrier because:
   unpark(t); <-- volatile read of t, can float up!

georges-gomes:
Interesting!

So there is potentially an issue on MCParkWaitStrat then.

SPParkWaitStrat is good with the volatile write right?

GG

georges-gomes:
I mean SCParkWaitStrat

georges-gomes:
Hi Nitsan, I pushed your comments.

Regarding the signaling.

##MCParkTakeStartegy
We are good because of the lock - ordering is safe
(I still need to figure out how to remove Condition if it generate garbage... Do you recommend we go for wait/notify?)

```java
    @Override
    public void signal()
    {
        ReentrantLock l = lock;
        l.lock();
        try
        {
            if (waiters>0)
            {
                cond.signal();
            }
        }
        finally
        {
            l.unlock();
        }
    }
```

##SCParkTakeStrategy
We are good with the volatile write and t.get() volatile read this will create a StoreLoad barrier right?

```java
    public volatile int storeFence = 0;

    @Override
    public void signal()
    {
        // Make sure the offer is visible before unpark
        storeFence = 1; // store barrier

        LockSupport.unpark(t.get()); // t.get() load barrier
    }
```

Let me know if you think something is not safe.

Cheers
GG

ChristianWulf:
According to http://shipilev.net/blog/2014/on-the-fence-with-dependencies/, it depends on the JVM implementation whether a write or a read to a volatile variable inserts a StoreLoad barrier.

nitsanw:
Ok, I'm going to merge this fucker :-)
But I'll be moving it to experimental and playing with it a bit before unleashing onto the core.

nitsanw:
merged!

georges-gomes:
May the force be with you :)
At your service if you need additional hands for some changes.
Cheers
GG

PR #56, opened 2015-03-19
Added profiles for benchmarks and experimental modules 

jitpack-io:
With this change `mvn install` will build the core module and 
`mvn install -Pbenchmarks,experimental` will build everything.


nitsanw:
Thanks!

PR #62, opened 2015-04-22
Queues rewritten with plain JDK atomics for platforms without Unsafe.

akarnokd:
I've converted most queue implementations. I've omitted class padding and array index scaling to save on memory for limited systems (such as Android).

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/atomic/AtomicReferenceArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/atomic/AtomicReferenceArrayQueue.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jctools.queues.atomic;
+
+import java.util.AbstractQueue;
+import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+
+import org.jctools.util.Pow2;
+
+abstract class AtomicReferenceArrayQueue<E> extends AbstractQueue<E> {
+    protected final AtomicReferenceArray<E> buffer;
+    protected final int mask;
+    public AtomicReferenceArrayQueue(int capacity) {
+        int actualCapacity = Pow2.roundToPowerOfTwo(capacity);
+        this.mask = actualCapacity - 1;
+        this.buffer = new AtomicReferenceArray<E>(actualCapacity);
+    }
+    @Override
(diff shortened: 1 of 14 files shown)


jctools-core/src/main/java/org/jctools/queues/atomic/AtomicReferenceArrayQueue.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+package org.jctools.queues.atomic;
+
+import java.util.AbstractQueue;
+import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+
+import org.jctools.util.Pow2;
+
+abstract class AtomicReferenceArrayQueue<E> extends AbstractQueue<E> {
```
nitsanw: We can save a dereference by extending AtomicReferenceArray an filling in the gaps in Queue methods. Not a must.

jctools-core/src/main/java/org/jctools/queues/atomic/MpscAtomicArrayQueue.java
```diff
@@ -0,0 +1,260 @@
...
+/**
+ * A Multi-Producer-Single-Consumer queue based on a {@link AtomicReferenceArrayQueue}. This implies that
+ * any thread may call the offer method, but only a single thread may call poll/peek for correctness to
+ * maintained. <br>
+ * This implementation follows patterns documented on the package level for False Sharing protection.<br>
+ * This implementation is using the <a href="http://sourceforge.net/projects/mc-fastflow/">Fast Flow</a>
+ * method for polling from the queue (with minor change to correctly publish the index) and an extension of
+ * the Leslie Lamport concurrent queue algorithm (originated by Martin Thompson) on the producer side.<br>
+ * 
+ * @author nitsanw
```
nitsanw: sometimes you attribute to me, sometimes you, sometimes both :)

nitsanw:
This looks good overall, I'll have a closer look and merge.
Did you benchmark?

nitsanw:
Also, thanks for contributing!!!

jctools-core/src/main/java/org/jctools/queues/atomic/AtomicReferenceArrayQueue.java
akarnokd: I thought about it but then we'd leak the public methods of AtomicReferenceArray into each queue implementation. Your suggestion only works in ```LinkedQueueAtomicNode``` only because it is not exposed.

nitsanw: Fine by me, I see your point.

jctools-core/src/main/java/org/jctools/queues/atomic/MpscAtomicArrayQueue.java
akarnokd: Due to copy-paste. We don't do attribution in RxJava so I usually forget about it.

akarnokd:
No benchmarks. I haven't yet familiarized myself with your benchmark subpackage.

nitsanw:
Oh, you'll have such fun :-)

nitsanw:
So, some provisional benchmarks show AtomicSpsc to have 6 times less throughput than Spsc. This is still 2.5 better than CLQ. I'm sure this can be improved on by re-introdcuing the counter padding, but will not persue unless someone shows an interest.

PR #67, opened 2015-05-12
unsafe methods as static-final

dorkbox:
There are a few unsafe methods that were not 'static final', and this pull request is to provide consistency with how the unsafe methods are called and to provide a hint at compile-time that the compiler should inline those methods.


nitsanw:
Thanks for the effort.
I don't think I'll be merging this in because the merge has gone beyond what I would see as needed here:
- It has a fair amount of formatting contents, removing end of line spaces. I prefer not to mix those with actual code changes.
- It changes the method signature, which is not required
- The signature change leads to a bunch of other changes, which are again not required.
I do agree with your motivation however and will fix appropriately.

dorkbox:
Great! Thanks for looking at it.

PR #68, opened 2015-05-12
Transfer array queue

dorkbox:
In Februrary, I found your MpmcArrayQueue, and was very impressed with the performance and lack of generated heap objects (for offer/poll). Being interested in having a similar capability to LinkedTransferQueue (but without any heap allocation), I modified your MpmcArrayQueue to create the "MpmcTransferArrayQueue", using a similar technique that Doug Lea used in the LinkedTransferQueue with regards to park/unpark rules. Since one of the requirements for your library is that it can run on java6, I did not depend on the TransferQueue interface -- I'm not sure the best way to resolve this, perhaps to have offer/poll block instead of having transfer/take?? I'm up for whatever suggestions you may have on this topic.

Here are my handrolled benchmarks for this queue (based on your spsc benchmark code), and I hope that this queue can find a home in your tools. The bechmarks were run on an i7-4700HQ CPU @ 2.40GHz w/ 16gigs of ram, running Linux 3.13.0-49-generic, x86_64. I did not pin the process to CPUs.

The non-blocking test results are based on your MpmcArrayQueue implementation, with a small (1024) queue size and a busy-spin on a full queue.

Blocking | LBQ | LTQ | MTAQ
----------------------|--------|-------|----------
1x1 Threads | 3,7m op/s | 6,3m op/s  | 3,6m op/s
2x2 Threads | 1,9m op/s | 2,4m op/s  | 2,9m op/s
3x3 Threads | 1,4m op/s | 1,8m op/s  | 2,0m op/s
4x4 Threads | 0,8m op/s | 1,4m op/s  | 1,5m op/s

Non-Blocking | LBQ | LTQ | MTAQ
----------------------|--------|-------|----------
1x1 Threads | 5,8m op/s | 8,1m op/s  | 30,7m op/s
2x2 Threads | 1,4m op/s | 3,4m op/s  | 6,4m op/s
3x3 Threads | 1,2m op/s | 2,3m op/s  | 3,8m op/s
4x4 Threads | 0,8m op/s | 1,8m op/s  | 2,4m op/s


nitsanw:
This looks interesting, I'll need some time to review the code so expect this to take a few days.
Thanks,
Nitsan

dorkbox:
@nitsanw It's been a while, is there any update on this?

jctools-core/src/main/java/org/jctools/queues/MpmcTransferArrayQueue.java
```diff
@@ -0,0 +1,613 @@
...
+}
+
+class Node<E> extends HI0<E> {
+    private static final long TYPE;
+    private static final long ITEM;
+    private static final long THREAD;
+
+    static {
+        try {
+            TYPE = UnsafeAccess.UNSAFE.objectFieldOffset(Node.class.getField("type"));
```
akarnokd: Why are these fields public? There is the `getDeclaredField()` to gain access to private fields.

dorkbox: No idea honestly -- I'll change those to be private.

nitsanw:
@dorkbox I'm sorry I didn't get back to you, life (work/children etc) got in the way and I had very little free time in the last few weeks. I hope to find the time to look at this shortly.

dorkbox:
@nitsanw No problem. Thanks for being able to look at it.

dorkbox:
@nitsanw  I finally had a chance to update and implement your suggestions for the transfer queue -- sorry this wasn't sooner, as a whole lot of things (work/family) got in the way. 

The Travis-ci subscription looks to be expired?? I'm not sure why it's failing, since I've tried several actions to resolve it, and the 'Details' link is non-resolvable for me.

I did notice some spurious annotations and number formatting (ie: 1_000_000 vs 1000000) in the benchmarks -- maybe that was causing issues? I corrected those, ci is still failing.

I'm all ears as to what might be going on.

dorkbox:
I just checked out jitpack -- and noticed it's failing. Fortunately it's more enlightening than travis-ci. I should have this updated soon.

dorkbox:
No idea why my IDE didn't pickup the missing imports -- very odd, but it's now passing.

dorkbox:
@nitsanw It's been a while, any update on this?

nitsanw:
Reviewing... please hold

nitsanw:
OK, Sorry for delaying this process for so long. Now reading through the code and the TrasferQueue interface/implementation on the JDK. At this stage I'm not even reviewing the correctness but trying to evaluate the usecase and how appropriate it is to offer your implementation a home in JCTools. I'm not convinced at first glance that this is a good idea. Which I realize is a disappointing result to arrive at after such a long wait. My reasoning so far (and I'm happy to hear your counter arguments):

 1. This is not an implementation of TransferQueue, as such it offers no 'drop-in' utility for people currently using TransferQueue. It has transfer/take, but that is only part of the full interface required. 
 2. The restriction on usage to allow either offer/poll OR transfer/take is confusing and error prone. This will not fail early and clearly, this will fail confusingly...

If this was a bounded TransferQueue implementation it would be a much better fit for JCTools.
I will continue to review the code and offer any advice I can on implementation, but this may have to live outside of JCTools for now. Would depending on JCTools and implementing this construct internally be an issue for you?

dorkbox:
No problem at all. Thank you for your suggestions and I appreciate the feedback.

 1. It could easily implement the TransferQueue interface, but that means a dependency on java 7+, which I'm not sure what your stance is, as your runtime is java 6. Having it fully implement the blocking queue interface isn't too difficult.

 2. I would need to write a compatible offer/poll (which was already written, but pulled out before I submitted the PR). Whatever guidance you have on this is welcome (as it's easy to put back). It's also easy to have it throw exceptions when those methods are called -- either way, I defer to your judgement.

Also, once the TransferQueue interface is implemented, this should qualify as a *legit* bounded TransferQueue, as it's already bounded and supports the park/unpark features required for transfer/take.  :)


Technically, it wouldn't be a problem to live internally in my projects since what I need access to for the implementation details have the 'protected' access modifier.

For the community, my hope was that this implementation could live with the similar utilities that you wrote, as I think that the visibility and functional cohesiveness is worth it. This is the **only** zero-GC, high performance, bounded, MPMC TransferQueue implementation that exists (as of May), excluding the Disruptor framework*.  

note: From my research (as of May), when the Disruptor queue is full it must completely empty before more data can be added to it; this is terrible for performance when there is no guarantee if the producers will be faster than the consumers. A key difference is that this queue implementation does not exhibit this specific problem.

nitsanw:
Further points:
 1. Please implement full TransferQueue interface. It will compile fine, but jdk6 user won't be able to use it and I can live with that. Please make offer/poll/transfer/put/take etc inter-operable. This is part of the motivation for this interface as I understand it.
 2. Please rename to MpmcArrayTransferQueue, making the interface the postfix as the convention goes
 3. Please use UnsafeRefArrayAccess utility for reference array access (you'll have to sync up with current code base, but should be straight forward), this should remove some of the boiler plate.
 4. For now, please move the implementation into the experimental module. This will allow a more gradual inclusion of this implementation.
 5. You refer to the Disruptor in your Javadoc, but the Disruptor doesn't really implement Queue/TransferQueue, so I'm not sure how the comparison is relevant. Am I missing something? If you are refering to the WorkerPool implementation please note that it is on it's way out of the Disruptor.

dorkbox:
Sure thing. I'll update this issue when those points are implemented.

For point 5, yes - it was in reference to the WorkerPool and how the Disruptor's producers/consumers operate (should one want to have functionality similar to a TransferQueue) and I didn't know it was on the way out - it's been a while.

nitsanw:
I've gone through the code and refactored/simplified to get my head around it. I've posted the refactored version as a gist:
 1. I removed the busy/yield/park bits, they should probably go back in when we're happy with the rest
 2. I renamed the node types to TAKE/TRANSFER
 3. I moved the node definition out of the class
 4. hasPendingMessages() was broken, looks like it got confused between hasConsumers and new meaning at some point.
 5. Documented assumption that queue is either empty or full of one type.
 6. Given 5 no need for consumer count
 7. Ordered element store into ref. array to correctly publish node type
 8. Clear node.item after node is taken to avoid memory leak.
See gist here, please comment: https://gist.github.com/nitsanw/7e67ef64ac89fd792839

dorkbox:
The changes look good, and have clarified how it works. Also removing consumer count is great!

I got sucked into work. I'll get back to this later today.

dorkbox:
I've got some initial work done to implement the TransferQueue interface (gist forked here: https://gist.github.com/dorkbox/c755d4b62645c350346e), however I'm not certain the best way to proceed with possible "bad/corrupt" indices as a result of tryTransfer (I suppose a solution would be to use a marker and have the next take/poll operation cleanup).

Previously, the queue was 'well structured' (either empty or all of a single type) and with the changes I just made [line 386](https://gist.github.com/dorkbox/c755d4b62645c350346e/revisions#diff-ec787c8c727e58cacb1ccba98442796cR386) changes the behaviour of this structure.

1) tryTransfer can corrupt/invalidate an index (if it fails to transfer), and the worst possible case is a single, blocked consumer and a producer that fills the queue, at which point the queue fails.

dorkbox:
@nitsanw I found an issue where the ['last item'](https://gist.github.com/nitsanw/7e67ef64ac89fd792839#file-refactoredmpmcatq-java-L194) can be changed out during the transfer/take methods. I suspect that this can happen when the queue loops around (so the pIndex and cIndex are still the same, but the contents at that index are different). 

This issue shows up very rarely and has been incredibly difficult to track down (even with additional sequence checks). I would say this is literally a show-stopper, and I haven't a clue how to correctly solve it.

If you want to continue down this path, I'm glad to continue working on it; otherwise I'm sorry for wasting your time, and you can close this PR, since even with this code as part of my own projects this is a serious enough issue that I cannot use the code in it's current state.

nitsanw:
Is this the original Homogenous TransferQueue or the mixed type version?

dorkbox:
This is in the mixed type version.

nitsanw:
OK, so at least we have a limited version which works. I'll give the mixed mode problem some thinking, see what I can come up with. I think a mixed mode could use a linked list of nodes (so add a prev. ref to the Node class) but ship those on the MPMC buffer. But its a very vague idea and I'd need to clear a few hours to look at this.

nitsanw:
@dorkbox So, in order to close the lid on this, can I ask you to cleanup your branch towards the changes suggested in comment (https://github.com/JCTools/JCTools/pull/68#issuecomment-167552876) and move the class into experimental.
I can then accept the PR in it's intermediary state and see if we can develop it further later.
WDYT?

dorkbox:
Sure. I'll move it to experimental and apply those changes. I'll also add source notes + caveats so this can be picked up later.

dorkbox:
@nitsanw life hit me going sideways (just had a kid), and I apologize for taking a month to get to something so trivial.

I added more comments, and refactored even more. Please let me know if this is acceptable since it is a [union of your initial refactor](https://gist.github.com/nitsanw/7e67ef64ac89fd792839) and what was there before. It has also been moved to experimental, the name was changed, and I added benchmarks (which are based on your handrolled ones) + comments about OSR, which is how I discovered the issue mentioned on [10 Jan](https://github.com/JCTools/JCTools/pull/68#issuecomment-170395307).

The benchmarks really should be moved to JMH, but I just haven't had the time.

nitsanw:
@dorkbox No problem! All good things come to those who wait. Congrats on having a kid, it involves a fair amount of mayhem in my experience, have fun.
I'll go over the PR and merge when I can.
Thanks again :-)

nitsanw:
@dorkbox I've had a go over the weekend, this is not looking too promising ATM.

* The code in it's current state does not run, I had to fix the class names for the unsafe field access to work.
* There are no tests. The existence of which would have helped highlight the above issue and perhaps others earlier on.
* To avoid confusion I reworked the hierarchy to not have this class implement Queue at all. I used the CircularArray class and added in the index fields.
* I removed the benchmarks (lots of code, many issues as you point out in comments. Upfront code compilation is not a real solution.). I wrote a simple JMH benchmark to compare this class and LTQ.
* On running the benchmark I've hit a bug where consumers and producers both block.

So overall, not a smooth merge. I might merge in the bits I got, but I don't know when I'll have time to crack the bug.

dorkbox:
@nitsanw That's not the best news, but I'm glad you were able to hit the issue where consumers and producers both block. That issue had bugged me for a while (and I am uncertain of the best way to solve it), which is why I switched my implementation (at the expense of simplicity) to use the Disruptor.  I'm definitely not an expert in this area, and if you want to close this merge request; that is OK with me. I fear that my attempt has gone down the wrong path.

nitsanw:
@dorkbox NP, I will sleep on it and see if I can work through the issue. It's an interesting problem in any case. Being an expert is just another way of describing the number of wrong paths you've been down, so keep at it :) these are slippery problems

dorkbox:
@nitsanw Sounds good. It is indeed slippery, and fortunately I've at learned a lot through-out this.

nitsanw:
Closing this PR, I don't have time to correct the algorithm. For reference I posted the fixed up version(compiles, does not extend Queue, other minor fixup) on gist:
https://gist.github.com/nitsanw/c713a7020cd753ea49d77f88ca4b40e5
Here's the simple benchmark I used to compare with LTQ:
https://gist.github.com/nitsanw/0b9f94bfd80a265a36ce93497f098382
When running with many consumers the benchmark hangs due to the aforementioned bug.
Thanks @dorkbox for all the work on this, perhaps we can get back to it in the future at some point.

PR #73, opened 2015-07-04
Added FastFlow-Lamport Mpsc ring buffer implementation + test

franz1981:
I added an implementation similar to the "original" Mpsc (FF + Lamport), in addition to the "plain" current Mpsc based only on FF.

### DIFF
--- a/jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
+++ b/jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
@@ -79,7 +79,7 @@ public class QueueBurstCost {
 
     }
 
-    static final Go GO = new Go();
+    Go GO = new Go();
 
     @Param({ "SpscArrayQueue", "MpscArrayQueue", "SpmcArrayQueue", "MpmcArrayQueue" })
     String qType;
@@ -136,7 +136,7 @@ public class QueueBurstCost {
     }
 
     @TearDown
-    public void killConsumer() throws InterruptedException {
+    public void killConsumers() throws InterruptedException {
         consumer.isRunning = false;
         for (int i = 0; i < consumerCount; i++) {
             consumerThreads[i].join();


franz1981:
Hi Nitsan, in the pull request did you need any kind of bench similar to those in your bench project or correctness tests?

nitsanw:
I could merge as is and fixup the bits I don't agree with, or you could do. I'll get round to it soonish(I hope)
Sorry to keep the PR waiting for so long.

franz1981:
Hi Nitsan, 
do not worry about the wait...my daily job did not make me feel it :) and in the meantime i'm (re)*studying THE CookBook because i feel i haven't understood it properly (what are the memory barriers issued and when :(). Thus, in 1 or 2 days i'll try to fix the sources using your suggestions but if for you is better to merge it as is and later fixing it (i don't want to waste your time reading a newbie code, considering that you already know what do you want to fix!) do it without problem :+1:
On the other hand if for you is the same i'm glad to fix it by myself and be corrected if i've not undestood you suggestions :)

franz1981:
Hi Nitsan, i've tried to update "my" implementation on what i've understand of your suggestions: the only dubt is about the original cas loop ...A repeated failed cas is really a rather rare case (more producers->more contention) but in order to allow the callers to choose it's own wait policy, i preferred to avoid the loop entirely: do you think is a good idea?

franz1981:
hi nitsan as soon as i can i'll submit a rewinded version of che writeAcquire (and read too!) without any WaitStrategy :) A little bit OT...in a past post with Moeller,he mentioned that the bounded queues are not well suited to be used as mailboxes for an a actor system...and a growable version is to be preferred...but why? The only reason i've found is to avoid in a simple manner the deadlock in case of cycles in the dependencies graph of the actors...but in this way way the backpressure will go away..!!

nitsanw:
the problem for actor systems with MANY queues is that the mailboxes for most actors are hardly ever back pressured. Preallocating them with large queues would limit the ability to support more actors due to memory utilization. The growable queues only grow when there's a buildup of messages in the queue.

franz1981:
Hi Nitsan!
Thank you for the answers, as ever :)
During the last months I was really (right) in the...job. As soon as i can i'll change and update my code to match the last versions of the API in a way to match the idea explained in http://psy-lob-saw.blogspot.it/2015/10/expanding-queue-interface-relaxed-queue.html
I'm glad that you wrote that article so with this update i'll kill two birds with one stone...

nitsanw:
I'm working through PRs, strike while the iron is hot and all that...

PR #79, opened 2015-09-28
offerIfBelowThreshold

ettoremaiorana:
This change is meant to add the ability for this queue to refuse additional items and apply back pressure on the producer 'before it's too late'.
The code is almost identical to offer(E e), with the exception of wrap point being the specified threshold. If threshold == queue-size then semantics match.


nitsanw:
Please add a test case for the new method as it is not part of the Queue/MessagePassingQueue API.
Thanks

ettoremaiorana:
Unit test added, let me know if you need a benchmark as well.

ettoremaiorana:
Thanks

PR #80, opened 2015-09-28
SpscUnboundedQueue leaves 1 slot null just before growing.

akarnokd:
In the following program, the queue allocates a new array even though
there is 1 slot available:

```java
SpscUnboundedArrayQueue<Integer> q = new SpscUnboundedArrayQueue<>(256);

for (int i = 0; i < 254; i++) {
    q.offer(i);
}

System.out.println(); // for breakpoint

q.offer(254);

System.out.println(); // for breakpoint
```

Clearly, the queue has room for 255 elements before it has to to grow.
Might not be much of an issue but for small queue sizes (minimum of 8
that the lookahead allows) is a relatively significant lost.


nitsanw:
Thank, I fixed the same issue on SpscGrowable a while back and forgot to do this one.

PR #85, opened 2016-01-28
Migration to Java 7, Maven tidied up and code inspections.

guidomedina:
The following has been done:
* Migrated to Java 7.
* Maven dependencies and plugins updated.
* Maven properties created at parent project and most things are now declared at parent project.
* IntelliJ 15.0.3 inspections and corrections applied with no side effect or functionality changed.

### DIFF
--- a/jctools-benchmarks/pom.xml
+++ b/jctools-benchmarks/pom.xml
@@ -1,70 +1,70 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
-	<modelVersion>4.0.0</modelVersion>
-	<parent>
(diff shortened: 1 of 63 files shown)


guidomedina:
Maven compiler bug starting at version 3.2 and up, fixing.

nitsanw:
Hi, JCTools core is Java 1.6, tests/experimental/benchmarks are 1.8.

guidomedina:
ok, will redo for Java 6 then, it is much nicer in Java 7 TBH, I don't get why people are stuck in Java 6 when there is basically free to move to Java 7 and gain lots of performance with new concurrent stuff, etc.

guidomedina:
It might be possible to create a new branch for Java 7 though? Some people might prefer the Java 7 version, you can always keep 1.1.x and 1.2.x and then 2.0 for Java 8? what do you think?

nitsanw:
:-) I'll drop Java6 support at some point I guess, but this is a product/lib so we have to accomodate the wider public. IT environments are slow to move and tend to have insane policies around these things made by people with a different decision making process...

PR #92, opened 2016-03-11
Implemented code generation for channel proxies.

raphw:
Works on my machine. There are still some todos with questions.

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/atomic/MpscAtomicArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/atomic/MpscAtomicArrayQueue.java
@@ -49,7 +49,7 @@ public final class MpscAtomicArrayQueue<E> extends AtomicReferenceArrayQueue<E>
      * concurrently.
      *
      * @see java.util.Queue#offer(java.lang.Object)
-     * @see MessagePassingQueue#offer(Object)
+     * @see org.jctools.queues.MessagePassingQueue#offer(Object)
      */
     @Override
     public boolean offer(final E e) {
@@ -130,7 +130,7 @@ public final class MpscAtomicArrayQueue<E> extends AtomicReferenceArrayQueue<E>
      * Lock free poll using ordered loads/stores. As class name suggests access is limited to a single thread.
      *
      * @see java.util.Queue#poll()
-     * @see MessagePassingQueue#poll()
+     * @see org.jctools.queues.MessagePassingQueue#poll()
      */
     @Override
     public E poll() {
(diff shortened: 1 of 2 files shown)


nitsanw:
@raphw sorry for taking long, I'm having a rather crowded week... I hope to get to reviewing and merging this in a few days

jctools-experimental/src/main/java/org/jctools/channels/proxy/DemoProxyResult.java
```diff
@@ -48,6 +49,7 @@ public int process(Demo1 impl, int limit) {
             case 2: {
                 float x = UnsafeAccess.UNSAFE.getFloat(rOffset + 4);
                 double y = UnsafeAccess.UNSAFE.getDouble(rOffset + 8);
+                // TODO: Should this offset not be +8 as the last value was a double?
```
nitsanw: yes, you are correct, should be:
boolean z = UnsafeAccess.UNSAFE.getBoolean(null, rOffset + 16);

jctools-experimental/src/main/java/org/jctools/channels/proxy/DemoProxyResult.java
```diff
@@ -81,6 +83,8 @@ public void call2(float x, double y, boolean z) {
         long wOffset = this.writeAcquire();
         UnsafeAccess.UNSAFE.putFloat(wOffset + 4, x);
         UnsafeAccess.UNSAFE.putDouble(wOffset + 8, y);
+        // TODO: Should this offset not be +8 as the last value was a double?
+        // TODO: What about object references, does that depend on compressed oops being enabled?
```
nitsanw: references require a parallel object references buffer.
You cannot (though some try) write refs into a byte array and reload them safely (because GC may move the ref).
You cannot (this one fails very quickly so I doubt people stick with it) write primitives into ref arrays because the JVM will try to treat these garbage pointers as pointers and crash.

nitsanw:
@raphw So, now this is the last remaining PR on the list (other one is experimental slow going one, blocked on the sender). Sorry I took so long, but here I am.

 1. Please send a separate PR for the JCTools shading, it's a completely split topic
 2. I was hoping for the code generation process (perhaps naively) to follow a different workflow where a dummy implementation is used as a template for the code generation. The dummy class would contain place holders for the generator to fill in, so that tweaking the template will be a plain Java editing effort. Is that possible?
3. I need to implement a parallel ref array for you to use from generated code.

nitsanw:
I've added some verification to the test, it seems the generated code is broken. Compared with the demo code (with fixed offset to 16 for the boolean and increased size):
test: https://gist.github.com/nitsanw/a9171e6b9ff5110b3917a12c9a6d3bc7
fixed demo code: https://gist.github.com/nitsanw/5d9683d800f6f63bc55a70c34324126d

raphw:
I am going to branch the two changes and revisit the generated code. I have to fix the offsets in the generated code and then I will run your test. There is surely a small error in there.

It is possible to template source code with a library like Javassist but I would recommend to stick with ASM in your case. I tried to arrange the code generation in a way that it is easier to understand and before mergin, I will add comments that indicate for each segment what they would represent in Java source code.

PR #94, opened 2016-03-14
Bring in latest Cliff's NHBM to JCTools

cliffclick:
Up-to-date copy of NonBlockingHashMap and friends.  JUnits for all tests, including performance tests.  No integration into maven, just the code and tests.


vyazelenko:
@cliffclick 
PR #94 is not Java 6 compatible. Travis build failed with compilation error (https://travis-ci.org/JCTools/JCTools/builds/115988521#L546):
```
[ERROR] /home/travis/build/JCTools/JCTools/jctools-core/src/main/java/org/cliffc/high_scale_lib/NonBlockingHashSet.java:[27,72] diamond operator is not supported in -source 1.6
```

cliffclick:
Cool, yeah I fought hard with maven & ideaj on what version of java was correct - "out of the box" i had all kinds of version-mismatch with JCTools.  :-(
I finally settled on java8 everywhere as a compromise.
I'll hack in java6 instead.
Cliff

vyazelenko:
This seems to do the trick, i.e. change line [NonBlockingHashSet.java#L27](https://github.com/cliffclick/JCTools/blob/c44fb009a2404a92b7c369b1f80425e32dfd3f8f/jctools-core/src/main/java/org/cliffc/high_scale_lib/NonBlockingHashSet.java#L27):
```diff
-  public NonBlockingHashSet() { super(); _map = new NonBlockingHashMap<>(); }
+  public NonBlockingHashSet() { super(); _map = new NonBlockingHashMap<E,Object>(); }
```

cliffclick:
Is the issue that it's 1.6 in the main code, and 1.8 in the test code?
When I ask IdeaJ to compile with 1.6, I get piles-o-errors in the test code.
Cliff

cliffclick:
Basically I'm busy unwinding all the java8-isms in the test code, and maybe I don't need to do that

vyazelenko:
Yes, the test code is allowed to use Java 1.8 but production code is only 1.6. Thus you can leave 1.8 in test code but keep production code 1.6.

cliffclick:
done!

vyazelenko:
The original problem with Intellij IDEA configuration was due to the fact that `jctools-core` module has both production code (1.6) and test code (1.8). However it is still a single `module` for IDEA which means that you must choose a single Java version for the entire module.

I guess that could be solved by either mandating that tests in `jctools-core` should also be 1.6 or by splitting out tests for `jctools-core` module as separate sub-project. But that's a general question about JCtools project structure to @nitsanw and @RichardWarburton.

vyazelenko:
@cliffclick NBHM tests are **not executed**, because they do not follow naming convention used by [maven-surefire-plugin](http://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html). However if I add this to the root `pom.xml` they will run:
```diff
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.19.1</version>
+                <configuration>
+                    <includes>
+                        <include>org/**/*.java</include>
+                    </includes>
+                </configuration>
+            </plugin>
```
Also the tests seems to be a mixture of performance tests/benchmarks and unit tests. Probably performance tests should be moved into `jctools-benchmarks` module.

nitsanw:
@cliffclick Thanks!!! This is a great contribution, I'll have a look at it as soon as I can.
@vyazelenko Thanks for helping :-)
The java8/6 split is mis-interpreted by the IDEs, both Eclipse and IDEA. This is a shame. Splitting out the tests seems like a bad idea to me, I think the maven build expresses the intent well at this stage. I think the confusion can be resolved through a contributor guide (very short, I'm not a tabs vs. spaces fanatic) clarifying the building and testing steps. What I do is setup Eclipse to be 1.8 and verify the code with maven before pushing.

nitsanw:
@cliffclick sorry for taking long, I'm having a rather crowded week... I hope to get to reviewing and merging this in a few days

nitsanw:
@cliffclick Working my way through, so far:

* Moved to experimental, this is temporary until we feel it's ready for release.
* Removed UnsafeUtil and replaced with UnsafeAccess
* Reviewed tests, I'll need to move the benchmarks to the benchmarks project and see about transforming them into JMH benchmarks where possible

Some questions:
* Why do we need NonBlockingHashtable? on a quick review it is just like NBHM, what am I missing?
* The licence quoted is CC, I'm not sure ow that plays with the APLv2 which JCTools is currently under. Would you mind if I moved it to APLv2? otherwise I would need some sort of compatible clause in the project licence, or something.

cliffclick:
On 4/10/2016 11:24 AM, Nitsan Wakart wrote:
>
> @cliffclick <https://github.com/cliffclick> Working my way through, so 
> far:
>
>   * Moved to experimental, this is temporary until we feel it's ready
>     for release.
>   * Removed UnsafeUtil and replaced with UnsafeAccess
>   * Reviewed tests, I'll need to move the benchmarks to the benchmarks
>     project and see about transforming them into JMH benchmarks where
>     possible
>
> Some questions:
>
>   * Why do we need NonBlockingHashtable? on a quick review it is just
>     like NBHM, what am I missing?
>
NBHtable was strictly a service utility, a drop-in replacements for 
Hashtables which had scaling problems.  I think the need for this has past.

>   * The licence quoted is CC, I'm not sure ow that plays with the
>     APLv2 which JCTools is currently under. Would you mind if I moved
>     it to APLv2? otherwise I would need some sort of compatible clause
>     in the project licence, or something.
>
APLv2 is fine.  I'll keep the existing SourceForge version under CC for 
those who care.

Cliff

nitsanw:
I've moved the handrolled benchmarks to benchmarks module and read through the lot.
I believe they could be replaced entirely with JMH benchmarks which would greatly simplify the them. I will tackle this separately however.
I've updated the licence where applicable and removed the Hashtable variant. 
I'll merge the branch in this state and handle the transition from experimental to core separately.

nitsanw:
Thanks again @cliffclick for all the work!
Please review the final state of the merged code and let me know if you got any comments.
The next steps as I see it:

 1. Invite review/comments from interested parties
 2. Port benchmarks to JMH, review results and see if there's any further wins to be had, or issues to fix
 3. Push to core and release.

cliffclick:
I'd like to be involved in that.  A number of my benchmarks stress-test 
non-obvious things, and I'm curious what is tested by the JMH 
benchmarks.  My first glance at them seemed that they missed a lot of 
stuff - especially around various kinds of heavy load testing.

Cliff

cliffclick:
Ok, here's my take on performance tests vs unit tests and "benchmarks".

If a test case runs on a 1000-core machine, running 1000 threads in parallel - and hits a single-threaded bottleneck, then the end performance is 1/1000th of what it should be.  That's a killer bug - and thus belongs as a ***test*** not a ***benchmark***.  A 1000x slowdown is just dead.  Been there, done that on Azul ~1000 core machines.  Best you get these days is a 64-core intel, but again - you should be able to get 64x speedups on 64 cores.  If you can't, then there's a bug in the performance of a supposed concurrent algorithm.  

I'd like to keep ***testing*** for (not ***benchmarking***) minimal parallel performance, under various kinds of odd-ball loads.  All reads, all writes, different blends, long-lived table contents, short-lived churning table contents, etc.

That being said, most of those tests don't have a reliable platform-independent way of confirming good scaling behavior as part of the test - typically because doing so requires the test platform have a known guaranteed level performance, and most test platforms don't deliver that.  i.e., if the machine is busy doing other random stuff, your performance varies so you can't check for e.g. linear scaling up to the CPU count.  Works great for individual developers, whose laptops are easily made "mostly idle" during the brief test run.

Cliff

nitsanw:
@cliffclick From reading the harness code (for CAT) it seems to accomplish several things:

 1. Separate warmup from results
 2. Format results, calculate stddev etc
 3. Spawn threads and synchronize start/stopping them
All of the above are achieved by JMH as well.
JMH comes with further niceties out the box, but it's most important feature is that it achieves good measurement infrastructure. It does so in a very similar way to the harness. That is at least my interpretation.
What JMH doesn't make an attempt at doing is guessing/determining the machine setup or allocated resources etc. It makes no attempt to "have a reliable platform-independent way of confirming good scaling behavior as part of the test", that is left to us fallible humans (to the mixed frustration and entertainment of our peers).
The harnesses, while wrapped in JUnit tests do not have a failing criteria. So do not assert any scaling properties.
At this point I feel that the harnesses can be expressed as JMH benchmarks without losing any functionality. The benchmarks measurements can then be run in various configurations to assert different properties and detect regressions (i.e. act as tests). 
I'll make sure I cc you on the conversion of the map tests and we can discuss in context.

cliffclick:
The CAT tests are the simplest.

The key take-away is that the harness has to add minimal overhead, or else the Counter isn't stressed, and doesn't ever need to auto-expand.  i.e., the JMH harness needs to be able to drive the CATCounter to the same levels of throughput as the CAT test harness.  If it can, and the CAT auto-expands out to the right size, you should get similar throughput as I get on big servers.  If the JMH harness adds too much overhead, the CAT counter won't need to expand and that behavior won't get tested.

As a separate useful function, the CAT test harness shows the throughputs of various other kinds of counting techniques.  While some counters do fine with low CPU counts, all start to choke up beyond 4 CPUs except the CAT counter - which should be testable with linear scaling up whatever you can get on an X86 these days (32 or 64 cores).  Also the harness also shows that if you don't have a race-safe counter you'll lose more than 99% of your counts.  I.e., the final counter value will effectively be random, and be uncorrelated with other racey counters - showing that racey counting is useless for high performance profiling work (the obvious use-case for racey counters).

Cliff

PR #96, opened 2016-03-15
SpscUnboundedArrayQueue prevent extended retention bw old and new gen

akarnokd:
I'm watching the Devoxx talk you linked and I saw SpscLinkedQueue already does some unreferencing.

I think the SpscUnboundedArrayQueue has the same issue (but with larger objects) where the consumer switches to the new buffer, the old buffer's next pointer isn't cleared.

In addition L199 should never be true because the new buffer receives the element before it becomes visible through the next of the old buffer.

This PR nulls out the next pointer in the old buffer for both safe and unsafe versions.


nitsanw:
Thanks, I'll include it in a minor release n the next few days.

PR #112, opened 2016-05-03
Implemented code generation for channel proxies.

raphw:
New PR on #92: This time with correct offset calculation (+8 after long and double). Added suggested test case.

### DIFF
--- a/jctools-experimental/src/main/java/org/jctools/channels/proxy/ProxyChannelFactory.java
+++ b/jctools-experimental/src/main/java/org/jctools/channels/proxy/ProxyChannelFactory.java
@@ -1,9 +1,424 @@
 package org.jctools.channels.proxy;
 
+import org.jctools.channels.spsc.SpscOffHeapFixedSizeRingBuffer;
+import org.jctools.util.UnsafeAccess;
+import org.objectweb.asm.*;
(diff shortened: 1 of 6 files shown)


jctools-experimental/src/main/java/org/jctools/channels/proxy/ProxyChannelFactory.java
```diff
@@ -1,9 +1,424 @@
...
+        methodVisitor.visitEnd();
+    }
+
+    private static void implementConstructor(ClassVisitor classVisitor) {
+        MethodVisitor methodVisitor = classVisitor.visitMethod(Opcodes.ACC_PUBLIC, "<init>", "(I)V", null, null);
+        methodVisitor.visitCode();
+
+        methodVisitor.visitVarInsn(Opcodes.ALOAD, 0);
+        methodVisitor.visitVarInsn(Opcodes.ILOAD, 1);
+        methodVisitor.visitIntInsn(Opcodes.BIPUSH, 13);
```
kay: Why is the messageSize hardcoded to 13? I see that the Pow2.align in the off heap ring buffer classes adjust this to 20.

nitsanw:
This is now merged via @kay PR which includes this one and further work.

Issue #108, opened 2016-04-19
Convert NHBM handrolled benchmarks to JMH benchmarks


nitsanw:

nitsanw:
@pcholakov has started some work here, please coordinate here if you want a piece of this

nitsanw:
Closing this as one of those has been done, will convert others at leisure.

PR #121, opened 2016-07-24
Add JMH benchmark for concurrent maps.

pcholakov:
This is pretty much a straight up port of the existing perf_hash_test to JMH, per #108. I moved SimpleRandom out of perf_hash_test.

Result of running both on the same machine:

```
jctools-benchmarks git:(nbhm-jmh)% for t in 2 4 6 8; do java -jar target/microbenchmarks.jar ConcurrentMapsThroughput.rand -t $t -f 1; done

# Threads: 2 threads, will synchronize iterations
# Benchmark: org.jctools.maps.nhbm_test.jmh.ConcurrentMapsThroughput.randomGetPutRemove
# Parameters: (implementation = NonBlockingHashMap, readRatio = 50, tableSize = 100000)

Benchmark                                      (implementation)  (readRatio)  (tableSize)   Mode  Cnt         Score         Error  Units
ConcurrentMapsThroughput.randomGetPutRemove  NonBlockingHashMap           50       100000  thrpt    6  17923933.571 ±  912895.637  ops/s
ConcurrentMapsThroughput.randomGetPutRemove   ConcurrentHashMap           50       100000  thrpt    6  23668440.241 ± 5684267.690  ops/s

###

# Threads: 4 threads, will synchronize iterations
# Benchmark: org.jctools.maps.nhbm_test.jmh.ConcurrentMapsThroughput.randomGetPutRemove
# Parameters: (implementation = NonBlockingHashMap, readRatio = 50, tableSize = 100000)

Benchmark                                      (implementation)  (readRatio)  (tableSize)   Mode  Cnt         Score         Error  Units
ConcurrentMapsThroughput.randomGetPutRemove  NonBlockingHashMap           50       100000  thrpt    6  44864770.568 ± 1202113.075  ops/s
ConcurrentMapsThroughput.randomGetPutRemove   ConcurrentHashMap           50       100000  thrpt    6  43945366.601 ± 3180110.185  ops/s

###

# Threads: 6 threads, will synchronize iterations
# Benchmark: org.jctools.maps.nhbm_test.jmh.ConcurrentMapsThroughput.randomGetPutRemove
# Parameters: (implementation = NonBlockingHashMap, readRatio = 50, tableSize = 100000)

Benchmark                                      (implementation)  (readRatio)  (tableSize)   Mode  Cnt         Score         Error  Units
ConcurrentMapsThroughput.randomGetPutRemove  NonBlockingHashMap           50       100000  thrpt    6  57554032.876 ±  335786.149  ops/s
ConcurrentMapsThroughput.randomGetPutRemove   ConcurrentHashMap           50       100000  thrpt    6  56251688.387 ± 3140116.465  ops/s

###

# Threads: 8 threads, will synchronize iterations
# Benchmark: org.jctools.maps.nhbm_test.jmh.ConcurrentMapsThroughput.randomGetPutRemove
# Parameters: (implementation = NonBlockingHashMap, readRatio = 50, tableSize = 100000)

Benchmark                                      (implementation)  (readRatio)  (tableSize)   Mode  Cnt         Score         Error  Units
ConcurrentMapsThroughput.randomGetPutRemove  NonBlockingHashMap           50       100000  thrpt    6  64094875.527 ± 2180806.338  ops/s
ConcurrentMapsThroughput.randomGetPutRemove   ConcurrentHashMap           50       100000  thrpt    6  62433923.436 ± 2661512.313  ops/s


jctools-benchmarks git:(nbhm-jmh)% java -cp ... org.jctools.maps.nhbm_test.perf_hash_test
50% gets, 25% inserts, 25% removes, table_size=100000 Best
Threads from 2 to 8 by 2
Warmup -variance:
===     CHM_16    2  cnts/sec=   22514871
===  NBHashMap    2  cnts/sec=   16893822 size=1048578
==== HashMap  Threads   Trial:    0          1          2          3          4          5          6          Avg      Stddev
===     CHM_16    2  cnts/sec=   17159725   24732722   24932298   25377222   23633198   25029976   25464857   24898332 (+/- 0%)  65377
===  NBHashMap    2  cnts/sec=   17281654   18891777   17008697   17564187   19901123   19047840   19239117   18501268 (+/- 4%)  65841 size=1048578
===     CHM_16    4  cnts/sec=   45679183   45948000   45803727   45934142   45873051   45820472   45274057   45967458 (+/- 0%)  65452
===  NBHashMap    4  cnts/sec=   45759998   46562757   46239087   46287162   46085904   45991481   46322006   46204051 (+/- 0%)  65443 size=1048578
===     CHM_16    6  cnts/sec=   57408141   58173032   58174207   58186218   58161588   58219358   58191284   58177819 (+/- 0%)  65507
===  NBHashMap    6  cnts/sec=   57648986   58911525   59107223   58633843   58790711   58970249   58758884   58820373 (+/- 0%)  65471 size=1048578
===     CHM_16    8  cnts/sec=   64764567   64139703   64566574   64774410   65508813   65302646   65318703   64947207 (+/- 0%)  65630
===  NBHashMap    8  cnts/sec=   65516278   63649737   65449784   66149245   65772432   65584908   65647868   65583018 (+/- 0%)  65458 size=1048578
```

Bonus: now supports HashMap as a baseline (this is not enabled by default). If run in a single-
threaded mode, the HashMap will *not* be synchronized.

### DIFF
--- a/jctools-benchmarks/src/main/java/org/jctools/maps/nhbm_test/SimpleRandom.java
+++ b/jctools-benchmarks/src/main/java/org/jctools/maps/nhbm_test/SimpleRandom.java
@@ -1,0 +1,26 @@
+package org.jctools.maps.nhbm_test;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+// Fairly fast random numbers
+public final class SimpleRandom {
+    private final static long multiplier = 0x5DEECE66DL;
+    private final static long addend = 0xBL;
+    private final static long mask = (1L << 48) - 1;
+    static final AtomicLong seq = new AtomicLong(-715159705);
+    private long seed;
+
+    public SimpleRandom() {
+        seed = System.nanoTime() + seq.getAndAdd(129);
+    }
+
+    public int nextInt() {
+        return next();
+    }
+
+    public int next() {
+        long nextseed = (seed * multiplier + addend) & mask;
+        seed = nextseed;
+        return ((int) (nextseed >>> 17)) & 0x7FFFFFFF;
+    }
+}
--- a/jctools-benchmarks/src/main/java/org/jctools/maps/nhbm_test/jmh/ConcurrentMapThroughput.java
+++ b/jctools-benchmarks/src/main/java/org/jctools/maps/nhbm_test/jmh/ConcurrentMapThroughput.java
@@ -1,0 +1,158 @@
+package org.jctools.maps.nhbm_test.jmh;
+
+import org.jctools.maps.NonBlockingHashMap;
+import org.jctools.maps.nhbm_test.SimpleRandom;
+import org.jctools.util.Pow2;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.ThreadParams;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Random;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+
+@OutputTimeUnit(TimeUnit.SECONDS)
+@BenchmarkMode({Mode.Throughput})
+@Warmup(iterations = 2, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 6, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
+@State(Scope.Benchmark)
+public class ConcurrentMapThroughput {
+
+    @Param(value = {"NonBlockingHashMap", "ConcurrentHashMap"})
+    private String implementation;
+
+    @Param(value = "50")
+    private static int readRatio;
+
+    @Param(value = "100000")
+    private static int tableSize;
+
+    private static String testData[];
+    private static int _gr, _pr;
+
+    private Map<String, String> map;
+
+    @Setup(Level.Trial)
+    public void createMap(ThreadParams threads) {
+        validateParameters();
+        createImplementation(threads);
+        setRatios();
+    }
+
+    private void validateParameters() {
+        if (readRatio < 0 || readRatio > 100) {
+            throw new IllegalArgumentException("readRatio must be a value between 0 and 100");
+        }
+        if (tableSize < 100 || tableSize > Pow2.MAX_POW2) {
+            throw new IllegalArgumentException("tableSize must be a value between 100 and " + Pow2.MAX_POW2);
+        }
+    }
+
+    private void createImplementation(ThreadParams threads) {
+        if ("ConcurrentHashMap".equalsIgnoreCase(implementation)) {
+            map = new ConcurrentHashMap<String, String>(16, 0.75f, 16);
+        } else if ("NonBlockingHashMap".equalsIgnoreCase(implementation)) {
+            map = new NonBlockingHashMap<String, String>();
+        } else if ("HashMap".equalsIgnoreCase(implementation)){
+            map = threads.getGroupCount() == 1 ?
+                    new HashMap<String, String>() : Collections.synchronizedMap(new HashMap<String, String>());
+        } else{
+            throw new IllegalArgumentException("Unsupported map: " + implementation);
+        }
+    }
+
+    private void setRatios() {
+        _gr = (readRatio << 20) / 100;
+        _pr = (((1 << 20) - _gr) >> 1) + _gr;
+    }
+
+    @Setup(Level.Trial)
+    public void prepareMap() {
+        testData = new String[Pow2.roundToPowerOfTwo(tableSize)];
+        for (int i = 0; i < testData.length; i++) {
+            testData[i] = String.valueOf(i) + "abc" + String.valueOf(i * 17 + 123);
+        }
+
+        Random rand = new Random();
+
+        int sz = map.size();
+        while (sz + 1024 < tableSize) {
+            int idx = rand.nextInt();
+            for (int i = 0; i < 1024; i++) {
+                String key = testData[idx & (testData.length - 1)];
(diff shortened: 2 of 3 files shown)


nitsanw:
This is great, thanks.
Looking at the data, I'm concerned with the differences in results, which may indicate a subtle difference in method.

```
* 2 Threads:
NonBlockingHashMap           50       100000  thrpt    6  17,923,933.571 ±  912,895.637  ops/s
ConcurrentHashMap            50       100000  thrpt    6  23,668,440.241 ± 5,684,267.690  ops/s
===     NBHashMap                          2  cnts/sec=   18,501,268 (+/- 4%)  65841 size=1048578
===        CHM_16                          2  cnts/sec=   24,898,332 (+/- 0%)  65377

* 4 Threads:
NonBlockingHashMap           50       100000  thrpt    6  44,864,770.568 ± 1,202,113.075  ops/s
ConcurrentHashMap            50       100000  thrpt    6  43,945,366.601 ± 3,180,110.185  ops/s
===  NBHashMap    4  cnts/sec=                            46,204,051 (+/- 0%)  65443 size=1048578
===     CHM_16    4  cnts/sec=                            45,967,458 (+/- 0%)  65452

6 Threads:
NonBlockingHashMap           50       100000  thrpt    6  57,554,032.876 ±   335,786.149  ops/s
ConcurrentHashMap            50       100000  thrpt    6  56,251,688.387 ± 3,140,116.465  ops/s
===  NBHashMap    6  cnts/sec=                            58,820,373 (+/- 0%)  65471 size=1048578
===     CHM_16    6  cnts/sec=                            58,177,819 (+/- 0%)  65507

8 Threads:
NonBlockingHashMap           50       100000  thrpt    6  64,094,875.527 ± 2,180,806.338  ops/s
ConcurrentHashMap            50       100000  thrpt    6  62,433,923.436 ± 2,661,512.313  ops/s
===  NBHashMap    8  cnts/sec=                            65,583,018 (+/- 0%)  65458 size=1048578
===     CHM_16    8  cnts/sec=                            64,947,207 (+/- 0%)  65630
```
* Original harness is able to squeeze more throughput from the benchmark, but this effect may be down to handrolled not synchronizing iterations (thus excerting less contention at beginning/end of measurement iterations).
* Results are similar, but the JMH results are very noisy, it would be good to repeat with a higher number of measurement iterations and forks.

If you have a large box to test on that would be great, I'll review the code shortly.

pcholakov:
Indeed, I also picked up on the increase in variance. I'll definitely test it on something bigger (and quieter than my laptop) but hadn't got to it yet. I'm seeing extra noise relative to the hand-rolled harness even for a 2-threaded JMH run so not entirely convinced I haven't subtly changed something - but if so, I don't know what, if you get a chance would appreciate an extra reviewer going through the code.

guidomedina:
Maybe this is something that could be also added to Caffeine's list of bench-marked caches and concurrent maps? as there it looks more informative:

https://github.com/ben-manes/caffeine/wiki/Benchmarks

All it takes there is to add the NBHM provider and that's it, since here you are only comparing to CHM there it will compare to other variants:

https://github.com/ben-manes/caffeine/wiki/Benchmarks

ben-manes:
fyi, NBHM will no-op on a `put` if the value references are equal. This short-circuit is a good optimization, but also makes it more difficult to construct a benchmark to show real-world performance. Because this causes write performance to have a large variance, a benchmark should take that into account or mention the difference with the results to reduce misinterpretations.

pcholakov:
Updated commit (sorry, I squashed it locally before pushing up):

- renamed benchmark class to ConcurrentMapThroughput
- minor refactoring of the setup method
- the benchmark method now returns the map keys it finds/updates/removes so the output can be sent to a blackhole consumer

[Trial run on a EC2 C4.8XL instance](https://gist.github.com/pcholakov/b8672ec2466a11c5834ab3e1677b9f22). Let me know if you'd like to see any other combinations. 

```
% numactl -N 1 -m 1 java -jar ... ConcurrentMap -t 18 -f 3 -tu us -p tableSize=10000000 -jvmArgs -Xmx16G -wi 10 -i 10
...
Benchmark                                     (implementation)  (readRatio)  (tableSize)   Mode  Cnt   Score   Error   Units
ConcurrentMapThroughput.randomGetPutRemove  NonBlockingHashMap           50     10000000  thrpt   30  43.778 ? 0.436  ops/us
ConcurrentMapThroughput.randomGetPutRemove   ConcurrentHashMap           50     10000000  thrpt   30  34.830 ? 4.643  ops/us 
```

pcholakov:
@ben-manes do you think a class-level comment about the NBHM `put` semantics would be sufficient?

ben-manes:
@pcholakov That should be fine. I have the following in my NBHM provider,
```
// Note that writes that update an entry to the same reference are short circuited
// and do not mutate the hash table. This makes those writes equal to a read.
```

If someone wants to use you benchmark to compare the two hash maps, hopefully they'll notice the note and not draw any hard conclusions without further investigation.

pcholakov:
Added (almost) the same comment, thanks Ben!

ben-manes:
lgtm

@nitsanw showed me a neat trick of using `@AuxCounters` to break down the operation into multiple metrics. That might be nice to show here, as raw throughput of one metric might mask a slowdown of another.

It might also be interesting to disable the `put` optimization to see how much of an effect it has.

You could also play with other distributions. I use Zipf since that matches real-world caches, which adds contention while improving cpu caching for hot entries. The current uniform might match application usage where there's a bunch of surrounding code, so there is a lower likelihood of contention or cpu caching playing much of a role.

pcholakov:
I was also looking at auxiliary counters initially but decided to keep it simple for this initial version. I wanted to limit the scope to just porting the original hand-rolled harness and see if we get the same numbers. The only deviation that will have a (smallish, constant) impact on performance numbers is the returning of values from the benchmark method which is a standard JMH idiom. Will be happy to add more real-world scenarios as benchmarks though, already have some alternatives locally.

nitsanw:
AuxCounters sounds like a good idea.
Please note the benchmarks use putIfAbsent and not put, so the comment while valuable doesn't apply to this benchmark.
I'm happy to take the PR as is and accept further improvements separately. Adding parametrized key distribution to the mix should be good, but can come later as it's not part of the porting effort.

ben-manes:
Sorry for my oversight, but the same general effect is in play. `NBHM` is optimistic and pre-screens the value before trying to update it. `CHM` is pessimistic by always acquiring the lock before making a decision. According to Doug this benefits of pre-screening cease at 32-cores where biased locking and safe-points disrupt the optimization. It is common for memoizers using `putIfAbsent` (or now `computeIfAbsent`) to pre-screen, so he opted to leave that as a user optimization. This is demonstrated in my compute benchmark.

My main point is that this difference makes it difficult to compare hash table algorithms to draw a conclusion of which is better and when.

nitsanw:
@ben-manes I take your point, and this benchmark is certainly lacking in certain ways. As a port of the existing coverage to JMH though it seems like fair representation.

ben-manes:
yep, as I said lgtm 👍

Issue #126, opened 2016-08-29
Support for OSGi desired or opposed?


CodingFabian:
Hi,
I just noticed that the jctools-core jar does not provide the required manifest entries, which would allow it to be used in osgi. Right now one would rebundle it to use it.
Are you guys opposed against adding some manifest entries? There is a maven plugin which does this automatically.
If you would accept a PR, I can have a look what would the best way for jctools to generate the manifest entries.

nitsanw:
There's no opposition, but I'm not clear on what is required for it to work.
If you send a PR that would be a great way to find out :)

nitsanw:
fixed with #127

PR #127, opened 2016-08-29
Generate OSGi Manifest during package phase

CodingFabian:
As discussed in #126
this will turn the maven packaging phase into making an OSGI bundle instead of a JAR file.
The only actual difference is that the felix bundle plugin analyzes imports and exports all classes.

The resulting manifest (build on my machine) looks like:
```
Manifest-Version: 1.0
Bnd-LastModified: 1472494973353
Build-Jdk: 1.8.0_71
Built-By: fabian
Bundle-Description: Java Concurrency Tools Core Library
Bundle-DocURL: https://github.com/JCTools
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
Bundle-ManifestVersion: 2
Bundle-Name: Java Concurrency Tools Core Library
Bundle-SymbolicName: org.jctools.core
Bundle-Version: 1.3.0.SNAPSHOT
Created-By: Apache Maven Bundle Plugin
Export-Package: org.jctools.queues.atomic;version="1.3.0";uses:="org.jct
 ools.queues,org.jctools.queues.spec",org.jctools.queues;version="1.3.0"
 ;uses:="org.jctools.queues.spec",org.jctools.queues.spec;version="1.3.0
 ",org.jctools.util;version="1.3.0";uses:="sun.misc"
Import-Package: sun.misc
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.6))"
Tool: Bnd-2.4.1.201501161923
```

It is worth noting that "sun.misc" is noted as required import, so this will not work on an j9 for example. I have not checked all references, but if it is handled gracefully (like catching class not found) then we should mark it as optional. 
Let me know if thats the case

PS: some whitespace changes done by my editor. i can remove them if desired.

### DIFF
--- a/jctools-core/pom.xml
+++ b/jctools-core/pom.xml
@@ -10,24 +10,35 @@
 	<artifactId>jctools-core</artifactId>
 	<name>Java Concurrency Tools Core Library</name>
 	<description>Java Concurrency Tools Core Library</description>
-	<packaging>jar</packaging>
-	
-    <dependencies>
-        <dependency>
-            <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-all</artifactId>
-            <version>${hamcrest.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>${junit.version}</version>
-            <scope>test</scope>
-        </dependency>
-    </dependencies>
+	<packaging>bundle</packaging>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.hamcrest</groupId>
+			<artifactId>hamcrest-all</artifactId>
+			<version>${hamcrest.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<version>${junit.version}</version>
+			<scope>test</scope>
+		</dependency>
+	</dependencies>
 	<build>
 		<plugins>
+			<plugin>
+				<groupId>org.apache.felix</groupId>
+				<artifactId>maven-bundle-plugin</artifactId>
+				<version>2.5.4</version>
+				<extensions>true</extensions>
+				<configuration>
+					<instructions>
+						<Import-Package>sun.misc;resolution:=optional</Import-Package>
+					</instructions>
+				</configuration>
+			</plugin>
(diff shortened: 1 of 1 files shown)


nitsanw:
Sorry for taking long to respond on this, currently travelling... Will review in the next week or so.
Thanks for doing the work!

CodingFabian:
I will probably have time to investigate the sun misc dependency until then. Enjoy JavaZone

CodingFabian:
ok, org.jctools.util.UnsafeAccess.SUPPORTS_GET_AND_SET somehow suggests that Unsafe is optional, but as far as I can see, Unsafe is very much required for jctools, so the generated manifest looks alright to me and I propose to merge it.
should there be activities which make unsafe optional, we can teach that the manifest generation.

normanmaurer:
@CodingFabian actually there are implementations that not need Unsafe. So I think it is not required for all implementations that are part of jctools.

CodingFabian:
But osgi declares this on package level. If desired,  the unsafe using and unsafe not using queues would need to be in separate packages.

CodingFabian:
We can already prep for the java 9 module madness :-)

CodingFabian:
i am going to declare it optional

CodingFabian:
Manifest now looks like this
```
Manifest-Version: 1.0
Bnd-LastModified: 1473413126204
Build-Jdk: 1.8.0_71
Built-By: fabian
Bundle-Description: Java Concurrency Tools Core Library
Bundle-DocURL: https://github.com/JCTools
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
Bundle-ManifestVersion: 2
Bundle-Name: Java Concurrency Tools Core Library
Bundle-SymbolicName: org.jctools.core
Bundle-Version: 1.3.0.SNAPSHOT
Created-By: Apache Maven Bundle Plugin
Export-Package: org.jctools.queues.atomic;version="1.3.0";uses:="org.jct
 ools.queues,org.jctools.queues.spec",org.jctools.queues;version="1.3.0"
 ;uses:="org.jctools.queues.spec",org.jctools.queues.spec;version="1.3.0
 ",org.jctools.util;version="1.3.0";uses:="sun.misc"
Import-Package: sun.misc;resolution:=optional
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.6))"
Tool: Bnd-2.4.1.201501161923
```

PR #129, opened 2016-10-05
Split Mpsc chunked into Mpsc LQ impls

nitsanw:
This is a split of Chunked into Unbounded/Chunked/Growable implementations with a unified base


nitsanw:
@normanmaurer please review impact on Netty. I intend to release it with 1.3.

PR #142, opened 2016-11-24
Channel proxies with references

kay:
Added support for calling methods with references. I kept the original ring buffer `OffHeapFixedMessageSizeRingBuffer` separate to `OffHeapFixedMessageSizeWithReferenceSupportRingBuffer` incase this experimental feature expanded for IPC. This may be wasted effort. The generalised protocol is that each call frame gets a fixed number of bytes for primitive arguments and a fixed number of indices in the ref array for ref arguments. This is fixed to be the largest required call frame for the time being. The ordering is the same as before, i.e. from `readAcquire`/`readRelease` and `writeAcquire`/`writeRelease`. The index to the reference arguments is calculated from a plain read of the producer or consume index within the acquire/release block.

In addition to implementing reference support a few notable bugs have been fixed

- The method `org.jctools.channels.proxy.ProxyChannelFactory.findExisting(String, Class<?>)` used the generated internal class name but the code paths here use the dotted separated class name
- The `messageSize`, that is the number of bytes required to write down all the primitive arguments of a call frame, was hard coded to 13 so I've added some logic that'll calculate that

Not really a bug but I've added _something_ (`WaitStrategy`) to handle what to do when the queue is full. I've not defaulted this so users must supply the strategy. Tests give example usage. This is implemented in `org.jctools.channels.spsc.SpscOffHeapFixedSizeWithReferenceSupportRingBuffer.writeAcquireWithWaitStrategy()` which I welcome some review of.

### DIFF
--- a/jctools-benchmarks/pom.xml
+++ b/jctools-benchmarks/pom.xml
@@ -36,7 +36,7 @@
             <artifactId>jmh-generator-annprocess</artifactId>
             <version>${jmh-core.version}</version>
             <scope>provided</scope>
-        </dependency>        
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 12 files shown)


nitsanw:
Looks good enough to merge into experimental. Thanks @kay and @raphw I'll play around with this and see where we go next.
Also thanks for your patience @raphw I should have merged this long ago.

raphw:
I just imagined you sitting on table mountain, meditating over the best approach.

Great this got finally merged. Always just ping me if you want a change or fix, I am happy to help maintain this.

PR #146, opened 2016-12-18
Replaced ordered release of message slot with a plain one.

franz1981:


pjulien:
This looks wrong to me.  The plain store can't stop tearing when 64-bit pointers are in use.

franz1981:
The original [MpscArrayQueue::poll](https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java#L323) and [MpscArrayQueue::poll](https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java#L380) have used the same operation to nullify the message slot value.
I'm hoping I've understood well your statement, but I can assume the tearing isn't an issue here, because the only reader of the value is the consumer itself; it is not like the FastFlow q that need to publish the value to another thread.

nitsanw:
OK, this looks good.
To clarify the discussion:
ONLY long/double can suffer word tearing. references are ALWAYS atomic. So using putOrdered for refs is all about ordering and not about atomicity.
The change is consistent with the poll, and looks good to me.
Thanks,
Nitsan

PR #149, opened 2016-12-23
MPSC/user custom channel + primitives only support.

kay:
Support for user specified back end data structures for the proxy byte
code generation factory. Users need only to specify a ProxyChannelRingBuffer
implementation. The protocol is to acquire/release for both loads and
stores. Within these borders plain stores and plain loads are performed.

Only subtle and implicit part of the API is that implementors of
ProxyChannelRingBuffer must provide the constructor (int capacity, int
messageSize, int referenceMessageSize).

This change supports the existing MPSC channel which I've added reference
support to. I've consolidated reference function from the class
OffHeapFixedMessageSizeWithReferenceSupportRingBuffer into the basic
class OffHeapFixedMessageSizeRingBuffer. When referenceMessageSize = 0
then the backing array will not be instantiated and nor will the reference
methods be supported. This is expressed as asserts rather than throwing an
exception with the expectation the lack of branch would perform better in
well behaved usage scenarios and NPE in others.

Extracted out the wait strategy to the org.jctools.channels package.
Modified how the wait strategy works as a field of the generated proxy
class and use a static util method on the factory for it's implementation,
thus letting the backend channels omit implementation.

Added to and heavily refactored tests to try to better separate and
express test intent.

Renamed arrayMessageSize to referenceMessageSize and messageSize to
primitiveMessageSize to create a consistent and easy to understand
distinction.

Moved the impl of ProxyChannelRingBuffer up to the top of
OffHeapFixedSizeRingBuffer. This may arguably be semantically ambiguous
but it seemed like the easiest thing to do. I am open to discussion of
removing the ProxyChannelRingBuffer interface and specifying it's API via
OffHeapFixedSizeRingBuffer with protected methods instead.

Added noddy MPSC just to give me an initial idea but it's not well thought
out. I plan to revisit both SPSC and MPSC benchmarks in order to compare
them to alternate solutions such as the disruptr, etc.


nitsanw:
Sorry for taking so long, merging!

PR #156, opened 2017-02-09
Multicast channel implementation

franz1981:
It includes 3 different flavours of the same data structure ie a single producer multicast channel:
a ring buffer that allows the producer to overwrite the published content without being slowed down by the consumers.
The consumers, like the LMAX Disruptor's handlers are able to read every produced message, but can join the transmission in any moment, reading the messages or simply chasing the latest produced message without reading it.
Each time a consumer will attempt to chase or join a transmission it could experience losses if not fast enough: in that case it will maintain the count of the overall lost messages.

The 3 flavours are:
1) A primitive (long) one (build as a base-line to be tested against the third flavour with LongValue flyweight)
2) An heap base one (maybe a freak experiment, considering that the messages will remain in the ring buffer until they will be overwritten)
3) An off-heap one (that will join the list of the off heap fixed size message channels)

Are provided sanity tests (that mimic the others of the already present data structures) and micro-benchmarks (hand-rolled and JMH based one).

The algorithm is based on the same idea of the FastFlow queue ie using an indicator for each message to store the state of the producer while approaching to the message slot (ready to be written/just written + how many times it is happened) but the consumers will read(-acquire) it just to verify if the message content is no longer valid and to know how many messages they have lost.

### DIFF
--- a/jctools-benchmarks/pom.xml
+++ b/jctools-benchmarks/pom.xml
@@ -13,6 +13,7 @@
 
     <properties>
         <jmh-core.version>1.14</jmh-core.version>
+        <java.version>1.8</java.version>
     </properties>
 
     <dependencies>
--- a/jctools-benchmarks/src/main/java/org/jctools/channels/multicast/SpOffHeapFixedMessageSizeAppenderTpt.java
+++ b/jctools-benchmarks/src/main/java/org/jctools/channels/multicast/SpOffHeapFixedMessageSizeAppenderTpt.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jctools.channels.multicast;
+
+import java.util.concurrent.TimeUnit;
+
+import org.jctools.util.JvmInfo;
+import org.jctools.util.UnsafeAccess;
+import org.jctools.util.UnsafeDirectByteBuffer;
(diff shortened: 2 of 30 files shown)


franz1981:
Considering how much code I've written and that there are parts I can remove...There is the need to make it shorter for a review?
As someone more clever than me said:"If I've had time I would have written a shorter letter" :)

nitsanw:
My apologies for letting this PR slide for so long. I have not had time for more than a passing glimpse. I am concerned with the API decisions made ("shout"?) but these can be tackled after the effect. I'll have another go at a review soon.

franz1981:
@nitsanw np! The concern around the API decisions are indeed well founded: it plays a big role around how much the data structures will be user-friendly and understandable by others, hence every feedback would be very appreciated :)
In addition, I'm concerned to have not found a proper way to fit the "multicast with losses" channel behaviour into the existing channel API: I'm feeling it is not good and intuitive as it could be...

franz1981:
@nitsanw I can take care of:
- move the off heap classes into their new home with the rest of the family
- cleanup the "less usefull" versions of the data structure ie on heap ones
- change the remaining names into something less stravagant

Re the deeper semantic changes (ie messages loss and multicasting) I can't say if it is a good match with the current Channels API, in particular for the `size`/`isEmpty` semantic.

IMO a multicast channel matches the general idea of being a channel per se and I'm more inclined to think that  `ChannelProducer::producedMessages`and `ChannelReceiver::receivedMessages` are a better match than `Channel::size`/`Channel::isEmpty` considering the nice separation of concerns of the producer/receiver APIs.
This consideration is not only related to this PR: it is something that I believe it worths a PR alone.

In summary: being a channel IMO just mean to have producer(s) and receiver(s) that can produce/receive messages on it, while being a monitorable (I'm sure other names could fit better than this one :P) channel extend its feature set with the ability of being queried for some(volatile) information like size/isEmpty (I would like just to drop'em just considering how is difficult to achieve them with a strict queue semantic ahah, just joking :P).
That's my 2 cents on that: a channel is not a queue (lucky us!) and I would be more happy to make it clear on the API too.

Probably this is the big question I need to get an answer first before forcing (and I won't never do that) the `multicast channel` abstraction to fit into the current `Channel` one.

Let me know what do you think about that Nitsan :)

nitsanw:
I don't think this can currently fit with JCTools supported APIs. I'm closing for now. Sorry :(

Issue #10, opened 2014-04-17
Setup a JCStress based test suite for queues


nitsanw:

nitsanw:
closing this issue, though of course we need more jcstress tests.

PR #158, opened 2017-03-23
add jcstress support for #10

victorparmar:
- created a new module `jctools-concurrency-test` for tests based on jcstress
- creating shaded jar
- execution instructions added to Readme
- wrote 3 tests for SpscArrayQueue 

Big thanks to @vyazelenko for helping with this MR


nitsanw:
AWESOME! thanks for doing this, there are ofcourse many more tests to write and run, I hope to see many more PRs now that it's set up.

PR #159, opened 2017-04-04
fix indentation

avalanche123:


jctools-core/src/main/java/org/jctools/queues/SpscArrayQueue.java
```diff
@@ -190,20 +190,20 @@ public final long lvConsumerIndex() {
...
-	public E relaxedPoll() {
-		return poll();
-	}
-
-	@Override
-	public E relaxedPeek() {
-		return peek();
-	}
+    @Override
+    public boolean relaxedOffer(E message) {
```
pcholakov: ``message`` could also be made final while you're at it

avalanche123:
@pcholakov done, thanks for your feedback!

nitsanw:
Looks good, thanks!

Issue #163, opened 2017-04-27
Exception Message clarity


Scottmitch:
The constructor on `SpscGrowableArrayQueue` [1][2] and `SpscChunkedArrayQueue.java` [3][4] validates argument bounds, but the exception message indicates a different limit. Should the exception message be updated or the range check?

[1] https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/SpscGrowableArrayQueue.java#L31
[2] https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/SpscGrowableArrayQueue.java#L35
[3] https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/SpscChunkedArrayQueue.java#L32
[4] https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/SpscChunkedArrayQueue.java#L36

nitsanw:
The limits need reviewing and a consistent handling be setup for the linked array constructors, both SPSC and MPSC

nitsanw:
Tests confirming behaviour need to be added

maseev:
@nitsanw You mean add tests that will check all these range checks?

nitsanw:
Yes, exercise the limits. Would also be sensible to extract these as constants, so future changes all converge to single definition of minimum size. It would be good to scan through the code base for min/max capacity validation and standardize the implementation/error messages etc.

maseev:
What do you think about [this](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/ObjectUtil.java) approach? So, instead of writing these checks yourself and throw exceptions:

```java
    private static int validateCapacity(int capacity) {
        if(capacity < 2)
            throw new IllegalArgumentException("Minimum size must be 2 or more");
        return capacity;
    }
```

we'll create some RangeUtil class that would do this kind of checks, so the above example could be rewritten like this:

```java
    private static int validateCapacity(int capacity) {
      return RangeUtil.checkGreaterThanOrEqual(capacity, 2, "capacity");
    }
```

nitsanw:
nice :)

PR #169, opened 2017-05-06
Implemented a consistent range handling approach

maseev:
Related to #163


nitsanw:
Thanks! Very much appreciated!

PR #177, opened 2017-05-13
SingleWriterHashSet improvements

qwwdfsad:
Here is review and multiple fixes/reproducers for SingleWriterHashSet:

1) Guava-testlib test added, fixed issue with broken iterator contract.
2) Iterator now loads values via `lvElement` instead of plain loads. Though I was not able to reproduce any issues with jcstress, I think it's a possible source of heisenbugs.
3) JMH test for concurrent maps is ported for sets, `SingleWriterHashSet`, `ConcurrentHashMap` and `NonBlockingHashSet` are tested. If you're interested I've got some results for c4.4xlarge amazon instance.
4) Added jcstress tests which reveal two problems with set compaction during removal.
Consider following scenario:
set: `[1a, 1b]`
Two threads, one calls `remove(1a)` and seconds calls `contains(1b)`.
Set state transitions are `[1a, 1b]` -> `[1b, 1b]` -> `[1b, null]`
Following problems are present:
Iterator may read same value twice.
`contains` call may return false even though element is present in set due to race: first slot is checked in first state and second slot is checked in third state.
I've thought how to solve second problem and still compact set on removes, but I'm not sure if it's possible at all. What I propose is to use tombstones and track both set size and effective size (size + tombstones) and resize when `effective size > resizeThreshold`. If you agree, I can fix it in separate PR (or as part of this one, but it will take some time)


nitsanw:
Thanks! this looks like a great contribution. It's going to take me some time to review it, but will probably merge soon.

JanecekPetr:
And if it was me, a random bystander, interested in the c4.4xlarge JMH results? By the way, PRs like these are what's amazing about open source software. What a valuable contibution!

jctools-experimental/src/main/java/org/jctools/sets/SingleWriterHashSet.java
```diff
@@ -268,9 +268,8 @@ public E next() {
         private void findNextVal() {
             E[] array = buffer;
             int i = nextValIndex;
-            E e = null;
             for (; i < array.length; i++) {
-                e = array[i];
+                E e = lvElement(array, calcElementOffset(i, array.length - 1));
```
nitsanw: I do not think  required. The buffer is read with a volatile load when the iterator is created. Future reads are ordered with that 'snapshot' read. The semantics here are not very clear for concurrent sets, technically giving you a copy to iterate on for reading is perfectly valid, and so is letting you see further changes to the set which happen concurrently.

nitsanw:
This is a good catch and a valid issue.
I think one solution would be to keep a mutation counter which can be mutated cheaply (lazySet) and checked before/after a contains operation. If the set has not been mutated you are fine, otherwise you'll need to repeat the op(if the element was not found). This has the benefit of being easy to implement and cheap in the low mutation case, but can have a terrible cost in the worst case.
The element can still 'disappear' from the view of iterators however.
Using tombstones was the first approach for this implementation. The downside is that performance is worse when you need to check for the tombstone case everywhere.

jctools-experimental/src/main/java/org/jctools/sets/SingleWriterHashSet.java
qwwdfsad: Good point, I will revert this change

qwwdfsad:
Yep, mutator counter may work (to be more specific, only 'removal' counter is required here) for set ops, but not for iterators.

I still think tombstones are better option:
1) Iterator will not be broken. With mutation token SWHS should explicitly state something like `iterator may not be valid during concurrent remove calls` which is kind a strange contract.
2) `contains` call will have consistent performance (and wait-free as a perk).
3) Implementation will be aligned with `NonBlockingHashMap` and with not yet existing `SingleWriterHashMap`


If you don't mind, I will rewrite SWHS to tombstones (within this branch) and compare it with previous implementation.

nitsanw:
@qwwdfsad To document the issue for my poor recollection:

- Iterator may return same element twice in current scheme. This issue will not be cured by a modification counter, but will go away in a TOMBSTONE implementation. I wonder if this is a strong contract for JDK implementations.
- From my experimentation, TOMBSTONE impl was slower, but that is probably fine.
- `contains` contract can be met with a mod counter. In any case a specialized implementation is valid such that it will not use an iterator.

qwwdfsad:
Sorry for that delay, was short on time and had problems with reproducing my own benchmarks.

I've implemented both mod count (SingleWriterHashSet2) and tombstones (SingleWriterHashSet) and played with various benchmark setups, here are most interesting ones:

For pure gets both implementations are more or less equal.
```
Benchmark                                 (implementation)  (readRatio)  (tableSize)   Mode  Cnt   Score   Error   Units
ConcurrentSetThroughput.rw:randomGet    NonBlockingHashSet           90       100000  thrpt   30  12.039 ? 0.985  ops/us
ConcurrentSetThroughput.rw:randomGet     ConcurrentHashSet           90       100000  thrpt   30  15.486 ? 1.064  ops/us
ConcurrentSetThroughput.rw:randomGet   SingleWriterHashSet           90       100000  thrpt   30  17.484 ? 1.260  ops/us
ConcurrentSetThroughput.rw:randomGet  SingleWriterHashSet2           90       100000  thrpt   30  17.386 ? 1.096  ops/us
```

For reads and spurious adds/removes in single thread mod count is faster.
```
ConcurrentSetThroughput.rw:randomContainsAddRemove    NonBlockingHashSet           90       100000  thrpt   30  11.395 ? 0.585  ops/us
ConcurrentSetThroughput.rw:randomContainsAddRemove     ConcurrentHashSet           90       100000  thrpt   30  13.877 ? 0.846  ops/us
ConcurrentSetThroughput.rw:randomContainsAddRemove   SingleWriterHashSet           90       100000  thrpt   30  13.478 ? 0.826  ops/us
ConcurrentSetThroughput.rw:randomContainsAddRemove  SingleWriterHashSet2           90       100000  thrpt   30  15.911 ? 1.054  ops/us
```

For mixed workload (one thread is calling randomContainsAddRemove and 4 threads are calling randomGet) mod count is still faster.

```
Benchmark                                               (implementation)  (readRatio)  (tableSize)   Mode  Cnt   Score   Error   Units
ConcurrentSetThroughput.rw                            NonBlockingHashSet           90       100000  thrpt   30  58.606 ? 3.611  ops/us
ConcurrentSetThroughput.rw:randomContainsAddRemove    NonBlockingHashSet           90       100000  thrpt   30  11.083 ? 1.724  ops/us
ConcurrentSetThroughput.rw:randomGet                  NonBlockingHashSet           90       100000  thrpt   30  47.522 ? 4.123  ops/us
ConcurrentSetThroughput.rw                             ConcurrentHashSet           90       100000  thrpt   30  72.042 ? 4.304  ops/us
ConcurrentSetThroughput.rw:randomContainsAddRemove     ConcurrentHashSet           90       100000  thrpt   30  11.490 ? 1.648  ops/us
ConcurrentSetThroughput.rw:randomGet                   ConcurrentHashSet           90       100000  thrpt   30  60.552 ? 4.366  ops/us
ConcurrentSetThroughput.rw                           SingleWriterHashSet           90       100000  thrpt   30  66.946 ? 4.337  ops/us
ConcurrentSetThroughput.rw:randomContainsAddRemove   SingleWriterHashSet           90       100000  thrpt   30  12.029 ? 1.453  ops/us
ConcurrentSetThroughput.rw:randomGet                 SingleWriterHashSet           90       100000  thrpt   30  54.917 ? 4.070  ops/us
ConcurrentSetThroughput.rw                          SingleWriterHashSet2           90       100000  thrpt   30  79.167 ? 4.026  ops/us
ConcurrentSetThroughput.rw:randomContainsAddRemove  SingleWriterHashSet2           90       100000  thrpt   30  14.296 ? 1.658  ops/us
ConcurrentSetThroughput.rw:randomGet                SingleWriterHashSet2           90       100000  thrpt   30  64.871 ? 4.040  ops/us
```

Generally tombstones are not that helpful: in most cases they degrade performance and can be replaced with either mod count impl or with ConcurrentHashSet.
SingleWriterHashSet is beneficial for reads-heavy workloads with rare removes, so it's fine to implement it using mod count (and fix somewhere in iterator behaviour).

I'll cleanup everything, re-check it again and contribute within a week.

nitsanw:
@qwwdfsad hi! I'm spending some time on JCTools this week, any chance you can finalize this PR?

qwwdfsad:
Sorry for the delay and thanks for a reminder, it was out of my head.
Found and committed exactly the changes I've benchmarked.

No javadoc is added, my public javadoc writing skills are far from perfect :(

Another thing which I found interesting:
In SWHS (and, after quick look, in other places) you are using fast-path/slow-path 'abstraction'.
Do you have any measurements of how it affects resulting data structures? Is it necessary?
I've tried to merge fast and slow paths in SWHS and was able to measure any difference in performance.
Why am I asking this: in other libraries (e.g. fastutil) everything comes through "slow path" with for-loop (not sure if for any reason though)

nitsanw:
Huge thanks!

PR #183, opened 2017-07-05
Add support of MessagePassingQueue for BaseSpscLinkedArrayQueue

franz1981:
Currently the BaseSpscLinkedArrayQueue's based queues do not implement the MessagePassingQueue interface.
I've provided a default (i.e. semantically equal to the MessagePassingQueue doc) implementation to allow polymorphic uses, tests and benchmarks using the enhanced API offered by MessagePassingQueue.
In future an optimized implementation of the same methods could be provided.
Just a note: I've collected all the default implementations in package private util class.


nitsanw:
Thanks, I'll have a look at the implementation shortly.

nitsanw:
Please add a test class and verify this passes the MPQ sanity tests

franz1981:
@nitsanw Done, I've added 3 MPQ tests for each concrete implementation

franz1981:
@I've replicated (with small changes in the initial size/capacity) the same sanity tests of the MpscLinkedArray cousins, but probably I could replace them with only one for the SpscLinkedArray childrens and removing the SpscChunkedMessagePassingQueueSanityTest::testMaxSizeQueue.
wdyt?

jctools-core/src/main/java/org/jctools/queues/MessagePassingQueueUtil.java
```diff
@@ -0,0 +1,91 @@
...
+                              MessagePassingQueue.Supplier<? extends E> s,
+                              int limit) {
+      int i = 0;
+      for (; i < limit && queue.relaxedOffer(s.get()); i++);
+      return i;
+   }
+
+   public static <E> int fill(MessagePassingQueue<? super E> queue, MessagePassingQueue.Supplier<? extends E> s) {
+      int i = 0;
+      while (queue.relaxedOffer(s.get())) {
```
nitsanw: This is broken.
If you "s.get()" there HAS to be room in the queue for the element, since you have no way of telling the supplier this particular element is not queued up.

jctools-core/src/main/java/org/jctools/queues/MessagePassingQueueUtil.java
```diff
@@ -0,0 +1,91 @@
...
+         idleCounter = 0;
+         c.accept(e);
+      }
+   }
+
+   public static <E> int fill(MessagePassingQueue<? super E> queue,
+                              MessagePassingQueue.Supplier<? extends E> s,
+                              int limit) {
+      int i = 0;
+      for (; i < limit && queue.relaxedOffer(s.get()); i++);
```
nitsanw: same issue

jctools-core/src/main/java/org/jctools/queues/MessagePassingQueueUtil.java
nitsanw: Would be good to add a test which demonstrates this issue to the MPQ Sanity

franz1981: You're right, sorry for the dumb error...hence it seems that ``fill`` can't be implemented using the ``relaxedOffer`` method (as it is).
I feel that I could use 2 different ways to deal with it:

1. implements ``fill`` in the queue
2. provide a relaxedOffer method that use a ``Supplier<? extends E> s``

I would like to do the first and remove the ``fill`` implementations from ``MessagePassingQueueUtil`` and add a test as you've suggested, wdyt?

nitsanw: Not a dumb error, I've made same error myself.  Should have added a warning to the javadoc and a test and so on.
I think the way to solve is to implement a fill(supplier, limit) in the base class, then use this method to implement the other 2 variants. A good default batch size is the chunk size.

franz1981: Thanks, so with the error a test has born: a good exchange :)

> I think the way to solve is to implement a fill(supplier, limit) in the base class, then use this method to implement the other 2 variants. A good default batch size is the chunk size.

I'll do it in that way then, agree :+1:

franz1981:
@nitsanw I've put a sanity test on the ``fill`` method (hoping it makes sense how I've done it) and a first implementation of the same.
Just one note: the implementation isn't optimized in the same way of the multi producer version that rightly is trying to minimize expensive multiple CAS.
Instead it tries to honor the given limit (batch size) as much as possible, because there is no contention with other producers. If my assumption is wrong in anyway I can try to build an optimized version around different assumptions of expected behaviours.

Issue #124, opened 2016-08-25
Implement iterator for Queues


mrniko:
https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/MpscChunkedArrayQueue.java#L145

Related to https://github.com/netty/netty/issues/5581

nitsanw:
Can you expand on why this is functionality you need from a queue?

mrniko:
I need to get all queue values in some cases.

nitsanw:
I understand you have no time to elaborate, sadly I am also short on time. Iterators have not thus far been a feature most users of JCTools can do without. Developing it comes at a price of time, effort, and maintenance. Unless you present a good case for it I don't think it will happen.

guidomedina:
@mrniko 

Most queues offer drain functionality but such thing can be emulated by consuming the queue hence if this is the case then there is no need for iterator.

An iterator for a circular, linked arrays and concurrent queue can be quite a challenge TBH.

nitsanw:
This is not super hard if we are willing to accept on consumer is allowed to iterate. You'll need an iterator similar to the drain loop, but without moving consumerIndex and shifting the buffers. Open to PRs.

nitsanw:
In the interest of enabling #60 it would be good to provide a limited iterator implementation such that:
 1. Remove for the tail works, but otherwise does not work.
 2. Iterator is assumed to have similar guarantees to size(), in the sense that concurrent modification can lead to inconsistent results.
 3. Iterator is an extended peek mechanism.

This is an easy picking for someone looking to cut their teeth, so PRs welcome, one queue at a time please.

Scottmitch:
Is there any preference to implementing this vs. just `remove` (limited to the consumer thread)? I'm interested in having the capability to remove from MPSC queues to prevent object references from being retained and prevent GC in some async use cases. The advantage of just implementing `remove` is avoiding iterator creation and having to worry about complexity of iteration vs. the more general usage provided by an Iterator.

nitsanw:
@Scottmitch How do you mean "just remove"? I assume you are refferring to `boolean remove(Object o)`
The issue with removing anything but the consumerIndex element (the next polled element) is that it leaves a gap in the underlying queue that needs handling. In a linked queue you can patch this gap, by dropping the node, but in an array backed data structure the `remove` will require another approach. Here's some thoughts on the matter (a cup of coffee worth, I'll charge it to your account ;)):
[...]

PR #184, opened 2017-07-31
Implement remove(Object) for single consumer linked atomic queues

Scottmitch:
Motivation:
Queue implementations in JCTools generally do not support removal of elements. This feature may be necessary to prevent object leaks by retaining references to objects in long lived queues.

Modifications:
- BaseLinkedAtomicQueue implements boolean remove(Object) API in accordance with the JDK's queue interface
- MpscLinkedAtomicQueue and SpscLinkedAtomicQueue override the necessary getNext method used by the remove method

Result:
Single consumer linked atomic queue implementations support remove(Object)

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
@@ -13,11 +13,11 @@
  */
 package org.jctools.queues.atomic;
 
+import org.jctools.queues.MessagePassingQueue;
+
 import java.util.AbstractQueue;
 import java.util.Iterator;
 import java.util.concurrent.atomic.AtomicReference;
-
-import org.jctools.queues.MessagePassingQueue;
 
 abstract class BaseLinkedAtomicQueue<E> extends AbstractQueue<E> {
     private final AtomicReference<LinkedQueueAtomicNode<E>> producerNode;
@@ -53,6 +53,55 @@ abstract class BaseLinkedAtomicQueue<E> extends AbstractQueue<E> {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * This method is only safe to call from the consumer thread!
+     */
+    @Override
+    public final boolean remove(Object o) {
+        if (null == o) {
+            return false; // Null elements are not permitted, so null will never be removed.
+        }
+
+        final LinkedQueueAtomicNode<E> originalConsumerNode = lpConsumerNode();
+        for (LinkedQueueAtomicNode<E> prevConsumerNode = originalConsumerNode, currConsumerNode = getNextConsumerNode(originalConsumerNode);
+             currConsumerNode != null;
+             prevConsumerNode = currConsumerNode, currConsumerNode = getNextConsumerNode(currConsumerNode)) {
+            if (o.equals(currConsumerNode.lpValue())) {
+                LinkedQueueAtomicNode<E> nextNode = getNextConsumerNode(currConsumerNode);
+                if (nextNode != null && prevConsumerNode != originalConsumerNode) {
+                    // We are removing an interior node.
+                    prevConsumerNode.soNext(nextNode);
+
+                    // Avoid GC nepotism because we are discarding the current node.
+                    currConsumerNode.soNext(null);
+                } else if (prevConsumerNode == originalConsumerNode) {
+                    // We are removing the consumer node.
+                    prevConsumerNode.soNext(null);
+                    spConsumerNode(currConsumerNode);
+                } else {
+                    // We are removing the producer node.
+                    prevConsumerNode.soNext(null);
+
+                    // If the producer node is pointing to the node we are removing then we should try to atomically
+                    // update the reference to move it to the previous node.
+                    if (!producerNode.compareAndSet(currConsumerNode, prevConsumerNode)) {
+                        // If the producer(s) have offered more items we need to remove the currConsumerNode link.
+                        while ((nextNode = currConsumerNode.lvNext()) == null);
+                        prevConsumerNode.soNext(nextNode);
+                    }
+
+                    // Avoid GC nepotism because we are discarding the current node.
+                    currConsumerNode.soNext(null);
+                }
+                currConsumerNode.spValue(null);
+
+                return true;
+            }
+        }
+        return false;
+    }
 
     @Override
     public String toString() {
@@ -95,6 +144,8 @@ abstract class BaseLinkedAtomicQueue<E> extends AbstractQueue<E> {
     public final boolean isEmpty() {
         return lvConsumerNode() == lvProducerNode();
     }
+
+    abstract LinkedQueueAtomicNode<E> getNextConsumerNode(LinkedQueueAtomicNode<E> currConsumerNode);
 
     protected E getSingleConsumerNodeValue(LinkedQueueAtomicNode<E> currConsumerNode, LinkedQueueAtomicNode<E> nextNode) {
         // we have to null out the value because we are going to hang on to the node
--- a/jctools-core/src/test/java/org/jctools/queues/atomic/ScLinkedAtomicQueueTest.java
+++ b/jctools-core/src/test/java/org/jctools/queues/atomic/ScLinkedAtomicQueueTest.java
@@ -1,0 +1,161 @@
+package org.jctools.queues.atomic;
+
+import org.junit.Test;
+
+import java.util.Queue;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public abstract class ScLinkedAtomicQueueTest {
+    protected abstract Queue<Integer> newQueue();
+
+    private static <T> void assertQueueEmpty(Queue<T> queue) {
+        assertNull(queue.peek());
+        assertNull(queue.poll());
+        assertTrue(queue.isEmpty());
+        assertEquals(0, queue.size());
+    }
+
+    private void removeSimple(int removeValue, int expectedFirst, int expectedSecond) throws InterruptedException {
+        final Queue<Integer> queue = newQueue();
+        Thread t = new Thread() {
+            @Override
+            public void run() {
+                queue.offer(1);
+                queue.offer(2);
+                queue.offer(3);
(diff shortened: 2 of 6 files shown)


Scottmitch:
@nitsanw - PTAL.

Related: https://github.com/JCTools/JCTools/issues/124#issuecomment-317850470

Scottmitch:
I'm guessing this approach can be extended to the unsafe linked variants too, but I wanted to get this approach reviewed before attempting to propagate the changes.

jctools-core/src/test/java/org/jctools/queues/atomic/ScLinkedAtomicQueueTest.java
```diff
@@ -0,0 +1,75 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public abstract class ScLinkedAtomicQueueTest {
+    protected abstract Queue<Integer> newQueue();
+
+    @Test
+    public void removeWorksInBeginMiddleEnd() throws InterruptedException {
```
Scottmitch: @nitsanw - I didn't integrate this with the typical `QueueSanityTest` and `ConcurrentQueueSpec` approach because not all queues support this feature yet and figured it would be better to keep independent for now. Let me know if you have a more preferred approach.

nitsanw:
Hi @Scottmitch !
Sorry for taking so long to review, will be looking at this shortly.
Thanks,
Nitsan

jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
```diff
@@ -53,6 +53,51 @@ protected final void spConsumerNode(LinkedQueueAtomicNode<E> node) {
...
+                        // If the producer(s) have offered more items we need to remove the currConsumerNode link.
+                        while ((nextNode = currConsumerNode.lvNext()) == null);
+                        currConsumerNode.soNext(currConsumerNode);
+                        prevConsumerNode.soNext(nextNode);
+                    }
+                } else {
+                    // We are removing the first node.
+                    currConsumerNode.spValue(null);
+                    spConsumerNode(currConsumerNode);
+                }
```
nitsanw: I think the logic could be simpler, but it can be reworked after accepted.

jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
```diff
@@ -53,6 +53,51 @@ protected final void spConsumerNode(LinkedQueueAtomicNode<E> node) {
...
+        }
+
+        for (LinkedQueueAtomicNode<E> prevConsumerNode = null, currConsumerNode = getNextConsumerNode(lpConsumerNode());
+             currConsumerNode != null;
+             prevConsumerNode = currConsumerNode, currConsumerNode = getNextConsumerNode(currConsumerNode)) {
+            if (o.equals(currConsumerNode.lpValue())) {
+                LinkedQueueAtomicNode<E> nextNode = getNextConsumerNode(currConsumerNode);
+                if (nextNode != null && prevConsumerNode != null) {
+                    // We are removing an interior node.
+                    prevConsumerNode.soNext(nextNode);
```
nitsanw: need to clear value and next to avoid GC nepotism

jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
```diff
@@ -53,6 +53,51 @@ protected final void spConsumerNode(LinkedQueueAtomicNode<E> node) {
...
+                } else if (prevConsumerNode == null && nextNode == null) {
+                    // We are removing the only node.
+                    currConsumerNode.spValue(null);
+                    spConsumerNode(currConsumerNode);
+                } else if (nextNode == null) {
+                    // We are removing the last node.
```
nitsanw: I prefer to use the terms consumer/producer nodes rather than last/first tail/head.

jctools-core/src/test/java/org/jctools/queues/atomic/ScLinkedAtomicQueueTest.java
nitsanw: That's fine, but I would prefer to have some simple tests on top of concurrent tests, so we can debug simple bugs simply.

jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
```diff
@@ -53,6 +53,51 @@ protected final void spConsumerNode(LinkedQueueAtomicNode<E> node) {
...
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * This method is only safe to call from the consumer thread!
+     */
+    @Override
+    public final boolean remove(Object o) {
+        if (null == o) {
```
Scottmitch: I used NPE because other methods which can throw or return `false` (e.g. `offer`) also take the "throw" approach. Both seem legit and I'm fine with which ever you prefer https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#remove-java.lang.Object-. FWIW I would prefer to avoid throwing and just return false too.

jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
Scottmitch: done

jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
Scottmitch: I consolidated to 3 blocks: interior, consumer, producer. However more simplification is welcome.

jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
```diff
@@ -53,6 +53,51 @@ protected final void spConsumerNode(LinkedQueueAtomicNode<E> node) {
...
+            throw new NullPointerException();
+        }
+
+        for (LinkedQueueAtomicNode<E> prevConsumerNode = null, currConsumerNode = getNextConsumerNode(lpConsumerNode());
```
Scottmitch: There wasn't any examples of iteration that I could find but the idea was to start `current` at a "real" node (that contains a value) and maintain `prev`/`next` to do the removal. `current` should always represent "real" node during iteration and is safe to compare the value. Do you have an alternative suggestion? I'm also fine if you want to re-work after the PR is pulled in too.

jctools-core/src/test/java/org/jctools/queues/atomic/ScLinkedAtomicQueueTest.java
Scottmitch: I added some simple tests too

Scottmitch:
@nitsanw - comments addressed PTAL.

jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
Scottmitch: (updated to return false)

jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
```diff
@@ -53,6 +53,57 @@ protected final void spConsumerNode(LinkedQueueAtomicNode<E> node) {
...
+                    // We are removing an interior node.
+                    prevConsumerNode.soNext(nextNode);
+
+                    // Avoid GC nepotism because we are discarding the current node.
+                    currConsumerNode.spValue(null);
+                    currConsumerNode.soNext(null);
+                } else if (prevConsumerNode == null) {
+                    // We are removing the consumer node.
+                    currConsumerNode.spValue(null);
+                    spConsumerNode(currConsumerNode);
```
nitsanw: At this point you have released the consumer node, but have not nulled out it's reference to next (currConsumerNode). So same nepotism issue.

I think the prevConsumerNode should initially be consumerNode, and not null;

nitsanw: On further thought, this would make the handling the same as first case, so reduces complexity further.

Scottmitch: > So same nepotism issue .... I think the prevConsumerNode should initially be consumerNode, and not null;

Good catch ... the "token" consumer node's next pointer was not being set to null. I'm not sure if the code can be the same as the case before, because the producer may have touched the `currConsumerNode.next`.

jctools-core/src/main/java/org/jctools/queues/atomic/BaseLinkedAtomicQueue.java
```diff
@@ -53,6 +53,55 @@ protected final void spConsumerNode(LinkedQueueAtomicNode<E> node) {
...
+                LinkedQueueAtomicNode<E> nextNode = getNextConsumerNode(currConsumerNode);
+                if (nextNode != null && prevConsumerNode != originalConsumerNode) {
+                    // We are removing an interior node.
+                    prevConsumerNode.soNext(nextNode);
+
+                    // Avoid GC nepotism because we are discarding the current node.
+                    currConsumerNode.soNext(null);
+                } else if (prevConsumerNode == originalConsumerNode) {
+                    // We are removing the consumer node.
+                    prevConsumerNode.soNext(null);
```
Scottmitch: in the `getSingleConsumerNodeValue` the `next` is set to itself instead of `null` ... is this necessary?

nitsanw:
Thanks for the contribution! and sorry for the slow turn around

nitsanw:
Upon further testing this approach fails for the SPSC case where remove is essentially making the queue MP. I will commit a revision shortly with a stress test which makes this manifest.
I don't have a solution to this ATM, but the this will still work with MPSC.

Scottmitch:
> remove is essentially making the queue MP

Yes this is true. To clarify, the issue is the lazySet on the producer thread may not be visible to the consumer thread by the time the CAS operation is done?

PR #186, opened 2017-08-09
Code gen for atomic queues with javaparser

kay:
Code generator that takes in a raw Java source file from the ArrayQueue
project and transforms it into output source code that will use the
builtin AtomicFieldUpdater's instead of unsafe.

This works for the major 4 variants, SPSC, SPMC, MPSC and MPMC.

The code is generated on each rebuild

I've committed it's output changes along with this generator.


jctools-core/src/main/java/org/jctools/queues/atomic/MpmcAtomicArrayQueue.java
```diff
@@ -92,31 +133,34 @@ public boolean offer(final E e) {
...
         int seqOffset;
         long seq;
-        long cIndex = Long.MIN_VALUE;// start with bogus value, hope we don't need it
+        // start with bogus value, hope we don't need it
+        long cIndex = Long.MIN_VALUE;
         do {
             pIndex = lvProducerIndex();
             seqOffset = calcSequenceOffset(pIndex, mask);
             seq = lvSequence(sBuffer, seqOffset);
-            if (seq < pIndex) { // consumer has not moved this seq forward, it's as last producer left
```
kay: Why is this comment stripped?

jctools-core/src/main/java/org/jctools/queues/atomic/MpmcAtomicArrayQueue.java
```diff
@@ -41,48 +46,84 @@ public final long lvProducerIndex() {
     }
 
     protected final boolean casProducerIndex(long expect, long newValue) {
-        return P_INDEX_UPDATER.compareAndSet(this, expect, newValue);
```
kay: Make these the same

jctools-core/src/main/java/org/jctools/queues/atomic/MpmcAtomicArrayQueue.java
```diff
@@ -41,48 +46,84 @@ public final long lvProducerIndex() {
...
+ * both arrays. We are trading memory to avoid false sharing(active and passive).
+ * <li>2 arrays instead of one: The algorithm requires an extra array of longs matching the size of the
+ * elements array. This is doubling/tripling the memory allocated for the buffer.
+ * <li>Power of 2 capacity: Actual elements buffer (and sequence buffer) is the closest power of 2 larger or
+ * equal to the requested capacity.
+ * </ol>
+ *
+ * @param <E>
+ *            type of the element stored in the {@link java.util.Queue}
+ */
```
kay: The generator should replace the jdoc with a note saying this is generated, link the generator class and the original class.

jctools-core/src/main/java/org/jctools/queues/atomic/MpmcAtomicArrayQueue.java
kay: This appears to be a bug in JavaParser caused when there's a comment on the following line to an if statement with a comment

jctools-core/src/main/java/org/jctools/queues/atomic/MpscAtomicArrayQueue.java
```diff
@@ -285,17 +291,17 @@ public final int failFastOffer(final E e) {
      * Lock free poll using ordered loads/stores. As class name suggests access is limited to a single thread.
      *
      * @see java.util.Queue#poll()
-     * @see org.jctools.queues.MessagePassingQueue#poll()
+     * @see MessagePassingQueue#poll()
      */
     @Override
     public E poll() {
-        final long cIndex = lvConsumerIndex(); // LoadLoad
```
kay: Another example of stripped comment

jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java
```diff
@@ -119,7 +123,7 @@ public MpmcArrayQueueL3Pad(int capacity) {
  * @param <E>
  *            type of the element stored in the {@link java.util.Queue}
  */
-public class MpmcArrayQueue<E> extends MpmcArrayQueueL3Pad<E> implements QueueProgressIndicators {
```
kay: Already implemented by a parent, so I removed it for simplicity

jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java
```diff
@@ -144,7 +148,8 @@ public boolean offer(final E e) {
             pIndex = lvProducerIndex();
             seqOffset = calcSequenceOffset(pIndex, mask);
             seq = lvSequence(sBuffer, seqOffset);
-            if (seq < pIndex) { // consumer has not moved this seq forward, it's as last producer left
```
kay: Moved this comment because of some bugs in javaparser meant this was a work around to preserve it

jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java
```diff
@@ -215,6 +220,7 @@ public E peek() {
         return e;
     }
 
+    // $gen:ignore
```
kay: In theory I could work out automatically to ignore these methods based on the intersection of the MessagePassingQueue and Queue interface. i.e. all those in  Queue but not in MessagePassingQueue are ignored but there are some special cases such as the MPSC offerIfBelowThreshold and failFastOffer which are not specified on any interface. 

Let's just keep this for now

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
```diff
@@ -89,8 +91,8 @@ protected final long lvProducerLimit() {
         return producerLimit;
     }
 
-    protected final void soProducerLimit(long v) {
```
kay: Renamed for consistency

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
```diff
@@ -103,6 +105,7 @@ public MpscArrayQueueL2Pad(int capacity) {
     }
 }
 
+//$gen:ordered-fields
```
kay: This tells the generator to strip all the static initialisers and fields from this class and replace them with atomic field updaters. It also tells them to replace the body of each method that looks like an ordered operation on a field with appropriate accessors on the atomic field updaters.

jctools-core/src/main/java/org/jctools/queues/SpmcArrayQueue.java
```diff
@@ -84,7 +86,7 @@ public final long lvConsumerIndex() {
         return consumerIndex;
     }
 
-    protected final boolean casHead(long expect, long newValue) {
```
kay: Renamed for consistency

jctools-core/src/main/java/org/jctools/queues/atomic/MpmcAtomicArrayQueue.java
```diff
@@ -13,20 +13,33 @@
...
+import static org.jctools.util.JvmInfo.CPUs;
 import org.jctools.util.RangeUtil;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+import java.util.concurrent.atomic.AtomicLongArray;
 
+/**
+ * NOTE: This class was automatically generated by org.jctools.queues.atomic.JavaParsingAtomicArrayQueueGenerator
+ * which can found in the jctools-build module. The original source file is MpmcArrayQueue.java.
+ */
```
kay: This automatically gets put on each class processed by the generator

jctools-core/src/main/java/org/jctools/queues/atomic/MpmcAtomicArrayQueue.java
```diff
@@ -72,17 +97,56 @@ protected final boolean casConsumerIndex(long expect, long newValue) {
...
     long p10, p11, p12, p13, p14, p15, p16, p17;
+
     public MpmcAtomicArrayQueueL3Pad(int capacity) {
         super(capacity);
     }
 }
 
+/**
+ * NOTE: This class was automatically generated by org.jctools.queues.atomic.JavaParsingAtomicArrayQueueGenerator
+ * which can found in the jctools-build module. The original source file is MpmcArrayQueue.java.
```
kay: The above is prepended to the original JavaDoc so that atomics variant also has this documentation easily accessible but a warning first that the class is generated.

jctools-core/src/main/java/org/jctools/queues/atomic/MpmcAtomicArrayQueue.java
```diff
@@ -72,17 +97,56 @@ protected final boolean casConsumerIndex(long expect, long newValue) {
...
+ * @param <E>
+ *            type of the element stored in the {@link java.util.Queue}
+ */
 public class MpmcAtomicArrayQueue<E> extends MpmcAtomicArrayQueueL3Pad<E> {
-    
-    public MpmcAtomicArrayQueue(int capacity) {
+
+    static final int RECOMENDED_POLL_BATCH = CPUs * 4;
+
+    public static final int RECOMENDED_OFFER_BATCH = CPUs * 4;
```
kay: These two static fields are used by the messaging passing method bodies. I haven't done anything to remove these. I don't know if I should as they are accessible by users?

jctools-core/src/main/java/org/jctools/queues/atomic/MpmcAtomicArrayQueue.java
```diff
@@ -92,31 +156,35 @@ public boolean offer(final E e) {
             throw new NullPointerException();
         }
         final int mask = this.mask;
-        final int capacity = mask + 1;
```
kay: This is an arbitary difference that retyping it to a long makes consistent with the original unsafe queues. Has no impact.

jctools-core/src/main/java/org/jctools/queues/atomic/MpmcAtomicArrayQueue.java
```diff
@@ -131,48 +199,49 @@ public E poll() {
...
-    
+
     @Override
     public E peek() {
         long cIndex;
         E e;
         do {
             cIndex = lvConsumerIndex();
             // other consumers may have grabbed the element, or queue might be empty
-            e = lpElement(calcElementOffset(cIndex));
```
kay: This was just missed when I homogenised the atomic queues, the generator has made it consistent.

jctools-core/src/main/java/org/jctools/queues/atomic/MpscAtomicArrayQueue.java
```diff
@@ -222,23 +266,12 @@ public boolean offer(final E e) {
...
-
-    /**
-     * A wait free alternative to offer which fails on CAS failure.
-     *
-     * @param e new element, not null
-     * @return 1 if next element cannot be filled, -1 if CAS failed, 0 if successful
-     * @deprecated This was renamed to {@link #failFastOffer(Object)} please migrate
-     */
-    @Deprecated
-    public final int weakOffer(final E e) {
```
kay: Moved to end of file, scroll down :)

jctools-core/src/main/java/org/jctools/queues/atomic/SpscAtomicArrayQueue.java
```diff
@@ -100,11 +142,11 @@ public SpscAtomicArrayQueueL3Pad(int capacity) {
  * 2012 - Junchang- BQueue- Efﬁcient and Practical Queuing.pdf <br>
  * </i> This implementation is wait free.
  *
- * @author akarnokd
+ * @author nitsanw
```
kay: Perhaps the generator should claim it's rightful glory?

PR #188, opened 2017-08-17
Split JvmInfo into unsafe and portable components.

kay:
Many classes access how many CPUs on a system which sadly also results in a call to unsafe during clinit which is not suitable for Android users.

See https://github.com/netty/netty/issues/7117#issuecomment-322972545 for example

Essentially all that's been done here is to split it into unsafe and portable variants. Manually tracing through the class hierarchy indicates to me this should be sufficient to ensure the 'atomic' queues are portable and resolve this issue.


nitsanw:
Thanks mate! Will cut a release later this weekend

PR #192, opened 2017-08-22
producerReferenceArrayIndex incorrect for multi producer use cases

kay:
Fixed bug whereby with MP proxy (or in the future, MC) you could
read a reference before it was written. This is due to writeAcquire()
advancing the producerIndex (or readAcquire() advancing the
consumerIndex in the MC) case.

When the producerReferenceArrayIndex() (or consumerReferenceArrayIndex())
method is called it reads the current producerIndex. Clearly sometimes
in the MP case this index will have changed between writeAcquire() and
producerReferenceArrayIndex() calls. Thus P0 who won the writeAcquire for
slot X, could end up skipping over the corresponding ref array slot if P1
completed writeAcquire() before P0 finished it's write to the ref array.
In such a case also P0 and P1 compete for P1's ref array slot, last write wins.

The result is nulls in some slots and incorrect values in others.

### DIFF
--- a/jctools-experimental/src/main/java/org/jctools/channels/OffHeapFixedMessageSizeRingBuffer.java
+++ b/jctools-experimental/src/main/java/org/jctools/channels/OffHeapFixedMessageSizeRingBuffer.java
@@ -33,10 +33,13 @@ import org.jctools.util.UnsafeRefArrayAccess;
  */
 public abstract class OffHeapFixedMessageSizeRingBuffer extends ProxyChannelRingBuffer {
 
-    public static final int READ_RELEASE_INDICATOR = 0;
-    public static final int READ_ACQUIRE_INDICATOR = 1;
-    public static final int WRITE_RELEASE_INDICATOR = 2;
-    public static final int WRITE_ACQUIRE_INDICATOR = 3;
+    /*
+     * Valid values in the buffer are >0, so we use <0 sentinel values for
+     */
+    public static final int READ_RELEASE_INDICATOR = -10;
+    public static final int READ_ACQUIRE_INDICATOR = -11;
+    public static final int WRITE_RELEASE_INDICATOR = -12;
+    public static final int WRITE_ACQUIRE_INDICATOR = -13;
     public static final byte MESSAGE_INDICATOR_SIZE = 4;
     public static final int HEADER_SIZE = 4 * CACHE_LINE_SIZE;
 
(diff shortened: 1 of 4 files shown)


jctools-experimental/src/main/java/org/jctools/channels/OffHeapFixedMessageSizeRingBuffer.java
```diff
@@ -33,10 +33,13 @@
  */
 public abstract class OffHeapFixedMessageSizeRingBuffer extends ProxyChannelRingBuffer {
 
-    public static final int READ_RELEASE_INDICATOR = 0;
-    public static final int READ_ACQUIRE_INDICATOR = 1;
-    public static final int WRITE_RELEASE_INDICATOR = 2;
-    public static final int WRITE_ACQUIRE_INDICATOR = 3;
+    /*
+     * Valid values in the buffer are >0, so we use <0 sentinel values for
+     */
```
kay: This isn't currently an issue with the existing SPSC and MPSC implementations but could easily be overlooked in other variations, so I've just preempted the issue. Perhaps unnecessarily but I don't see the harm.

jctools-experimental/src/main/java/org/jctools/channels/proxy/ProxyChannelFactory.java
```diff
@@ -181,7 +183,7 @@ public static long writeAcquireWithWaitStrategy(ProxyChannelRingBuffer channelBa
     }
 
     private static void implementUserMethods(ClassWriter classWriter, List<Method> relevantMethods, String generatedName, Class<?extends ProxyChannelRingBuffer> backendType) {
-        int type = 1;
+        int type = START_TYPE_ID;
```
kay: Constant to set the start of the callTypeId range

jctools-experimental/src/main/java/org/jctools/channels/proxy/ProxyChannelFactory.java
```diff
@@ -271,7 +278,12 @@ private static void implementProcess(ClassVisitor classVisitor,
 
         // switch(UnsafeAccess.UNSAFE.getInt(rOffset)) // start with case 1, increment by 1; represents "type"
         getUnsafe(methodVisitor, int.class, localIndexOfROffset, 0);
-        methodVisitor.visitTableSwitchInsn(1, cases.length, endOfSwitch, cases);
+        methodVisitor.visitVarInsn(Opcodes.ISTORE, localIndexOfTypeId);
+        methodVisitor.visitVarInsn(Opcodes.ILOAD, localIndexOfTypeId);
```
kay: Assign the result of UnsafeAccess.UNSAFE.getInt(rOffset) to a local variable to aid debugging of the generated proxy

jctools-experimental/src/main/java/org/jctools/channels/proxy/ProxyChannelFactory.java
```diff
@@ -286,8 +298,9 @@ private static void implementProcess(ClassVisitor classVisitor,
             int localIndexOfArrayReferenceBaseIndex = Integer.MIN_VALUE;
             for (Class<?> parameterType : method.getParameterTypes()) {
                 if (!parameterType.isPrimitive()) {
-                    // long referenceArrayIndex = this.consumerReferenceArrayIndex();
-                    consumerReferenceArrayIndex(methodVisitor, backendType);
+                    // long referenceArrayIndex = this.consumerReferenceArrayIndex(rOffset);
+                    consumerReferenceArrayIndex(methodVisitor, localIndexOfROffset, backendType);
+                    // TODO: Should reuse local indices or when an iFace grows too big we might have problems here
```
kay: While rereading this I noted that there's a risk of the locals table getting too big if we start proxying interfaces with lots of methods using references. e.g.

interface Alice {
  void f(Bob b);
  void g(Charlie c, David d);
}

Would result in 3 entries allocated in the locals table. Clearly this isn't necessary as the most needed is in the case of g(Charlie, David) which needs 2. This is due to the locals util being scoped for the entire method. Note an issue for small interfaces or within the context of this bug fix but I did add a comment to highlight it for the future.

jctools-experimental/src/main/java/org/jctools/channels/proxy/ProxyChannelFactory.java
```diff
@@ -342,7 +355,20 @@ private static void implementProcess(ClassVisitor classVisitor,
         // return i;
         methodVisitor.visitVarInsn(Opcodes.ILOAD, localIndexOfLoopIndex);
         methodVisitor.visitInsn(Opcodes.IRETURN);
+        
 
+        // Label the end of scope for all locals
+        Label localScopeEnd = new Label();
+        methodVisitor.visitLabel(localScopeEnd);
+
+        // Declare local variables to aid in debugging
```
kay: Self explanatory. This names the locals in the debugger variables page when looking at this generated frame.

nitsanw:
Where is the unit test to demonstrate the bug?

PR #195, opened 2017-09-14
XADD MultiProducer SingleConsumer Array Queue

franz1981:
The mechanics of this queue is similar on how Aeron's Publication (aka the LogBuffer) works, but using only 2 partitions and without background cleaning of the consumed elements.
When hit by a single producer the offer and poll has the same semantics of the MpscArrayQueue, otherwise there is the chance that more than the configured capacity will be written (without element losses), exactly as the Aeron's LogBuffer.
Benchmarks shows (throughput-wise) perf similar to MpscArrayQueue with Single Producer and a better scalability otherwise (mainly due to XADD).


franz1981:
@nitsanw The name Relaxed comes from the backpressure model that became more relaxed with more producers involved.

jctools-core/src/main/java/org/jctools/queues/MpscRelaxedArrayQueue.java
```diff
@@ -0,0 +1,686 @@
...
+    {
+        UNSAFE.putOrderedLong(this, C_POS_OFFSET, newValue);
+    }
+}
+
+/**
+ * NOTE: This class was automatically generated by org.jctools.queues.atomic.JavaParsingAtomicArrayQueueGenerator
+ * which can found in the jctools-build module. The original source file is MpscArrayQueue.java.
+ */
+abstract class MpscAtomicArrayQueueL3Pad<E> extends MpscAtomicArrayQueueConsumerPositionField<E>
```
nitsanw: You are mixing Atomic and Unsafe naming/implementation

jctools-core/src/main/java/org/jctools/queues/MpscRelaxedArrayQueue.java
```diff
@@ -0,0 +1,686 @@
...
+        while (poll() != null)
+        {
+            // if you stare into the void
+        }
+    }
+
+    @Override
+    public boolean isEmpty()
+    {
+        return size() == 0;
```
nitsanw: Should be:
```
       final long producerPosition = producerPosition(
            cycleId(producerCycleClaim, this.cycleIdBitShift),
            positionOnCycle(producerCycleClaim, this.positionOnCycleMask),
            this.cycleLengthLog2);
        return (producerPosition == consumerPosition);
```
?

nitsanw:
Thanks, this is a valuable port, I'll read carefully.
I'm not sure about the naming, but this can be worked out later.

jctools-core/src/main/java/org/jctools/queues/MpscRelaxedArrayQueue.java
```diff
@@ -0,0 +1,686 @@
...
+    private final int cycleLengthLog2;
+    private final E[] buffer;
+    private final long[] producerCycleClaim;
+    private final int positionOnCycleMask;
+    private final int cycleIdBitShift;
+
+    public MpscRelaxedArrayQueue(int capacity)
+    {
+        this.buffer = CircularArrayOffsetCalculator.allocate(Pow2.roundToPowerOfTwo(capacity) * 2);
+        this.producerCycleClaim = UnsafeLongArrayAccess.allocate(2);
```
nitsanw: since there's only 2 slots you could implement this array as 2 inlined fields, but still access them as an array (using Unsafe for arithmatic)

franz1981: :+1: TBH I've already tried it but using an `if` to choose the field: probably that hasn't exploit the improved locality..wdyt? if it worths I would be happy to retry it using math to access them...

franz1981:
@nitsanw 

> I'm not sure about the naming

I'm always so bad to name things argh

jctools-core/src/main/java/org/jctools/queues/MpscRelaxedArrayQueue.java
franz1981: good point!!!

jctools-core/src/main/java/org/jctools/queues/MpscRelaxedArrayQueue.java
mjpt777: It is worth avoiding the data dependant load especially in a larger application where it can cause a stall. In a microbenchmark it is less likely to show up.

franz1981: Hence it is not simple to show up the improvement of that in a microbench...and probably not on throughput based ones....

nitsanw: It's reasonable to tackle as a separate improvement in any case

nitsanw:
@mjpt777 AFAIK this is your algorithm, and you are a passionate namer, do you have a suggestion for a name to reflect the algorithm?
MpscDoubleBufferedQueue? MpscThompsonQueue? MpscLogBufferQueue?

franz1981:
@mjpt777 @nitsanw 

> MpscThompsonQueue

:100:  I'm not aware of any other q (or ring buffer) with a similar mechanics, hence it is well deserved :+1:

jctools-core/src/main/java/org/jctools/queues/MpscRelaxedArrayQueue.java
franz1981: Implementing it like this in the atomic version of the q:
```
    @Override
    public boolean isEmpty()
    {
        final long consumerPosition = lvConsumerPosition();
        final int activeCycleIndex = lvActiveCycleIndex();
        final long producerCycleClaim = AtomicLongArrayAccess.lvValue(this.producerCycleClaim, activeCycleIndex);
        final long producerPosition = producerPosition(
            cycleId(producerCycleClaim, this.cycleIdBitShift),
            positionOnCycle(producerCycleClaim, this.positionOnCycleMask),
            this.cycleLengthLog2);
        return (producerPosition == consumerPosition);
    }
```
It will cause a fail int the sanity test:
```
java.lang.AssertionError: Observed no element in non-empty queue 
Expected :0
Actual   :350
 <Click to see difference>


	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.jctools.queues.QueueSanityTest.testPollAfterIsEmpty(QueueSanityTest.java:361)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:51)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
```

franz1981: @nitsanw 
Turning it into this seems to fix that:
```
    @Override
    public boolean isEmpty()
    {
        final AtomicLongArray producerCycleClaims = this.producerCycleClaim;
        final int cycleLength = this.cycleLength;
        final int positionOnCycleMask = this.positionOnCycleMask;
        final int cycleIdBitShift = this.cycleIdBitShift;
        final int cycleLengthLog2 = this.cycleLengthLog2;
        final long consumerPosition = lvConsumerPosition();
        int activeCycleIndex;
        long producerCycleClaim;
        int positionOnCycle;
        do
        {
            activeCycleIndex = lvActiveCycleIndex();
            producerCycleClaim = AtomicLongArrayAccess.lvValue(producerCycleClaims, activeCycleIndex);
        }
        while ((positionOnCycle = positionOnCycle(producerCycleClaim, positionOnCycleMask)) > cycleLength);

        final long producerPosition = producerPosition(
            cycleId(producerCycleClaim, cycleIdBitShift),
            positionOnCycle,
            cycleLengthLog2);
        return (producerPosition == consumerPosition);
    }
```
Makes sense?
But TBH I'm not seeing such a big improvement vs the original implementation...

franz1981:
@mjpt777 @nitsanw I've added a second commit with the producer claim fields inlined into the q class.
If it is implemented in a safe/right way I will squash it in the first commit

franz1981:
I've modified the benchmarks of https://github.com/real-logic/benchmarks to run the `ManyToOneConcurrentArrayQueueBenchmark` bench (`burst length = 100`) vs MpscRelaxedArrayQueue/MpscArrayQueue using a relaxedPoll (to match the LogBuffer consume semantic) and this is what I'm getting:
```
Benchmark                                                             (burstLength)    Mode     Cnt       Score     Error  Units
AeronIpcBenchmark.test1Producer                                                 100  sample  116035    5602.900 ± 126.159  ns/op
AeronIpcBenchmark.test1Producer:test1Producer·p0.99                             100  sample           24160.000            ns/op
AeronIpcBenchmark.test2Producers                                                100  sample  219839   11341.917 ±  66.733  ns/op
AeronIpcBenchmark.test2Producers:test2Producers·p0.99                           100  sample           38848.000            ns/op
AeronIpcBenchmark.test3Producers                                                100  sample  247221   15645.017 ± 232.384  ns/op
AeronIpcBenchmark.test3Producers:test3Producers·p0.99                           100  sample           49472.000            ns/op

MpscArrayQueueBenchmark.test1Producer                                           100  sample  161740    3906.104 ±  16.878  ns/op
MpscArrayQueueBenchmark.test1Producer:test1Producer·p0.99                       100  sample           17216.000            ns/op
MpscArrayQueueBenchmark.test2Producers                                          100  sample  223930   11185.586 ±  23.686  ns/op
MpscArrayQueueBenchmark.test2Producers:test2Producers·p0.99                     100  sample           27776.000            ns/op
MpscArrayQueueBenchmark.test3Producers                                          100  sample  310557   24080.455 ±  18.060  ns/op
MpscArrayQueueBenchmark.test3Producers:test3Producers·p0.99                     100  sample           30144.000            ns/op

MpscRelaxedArrayQueueBenchmark.test1Producer                                    100  sample  139179    4527.424 ±  18.276  ns/op
MpscRelaxedArrayQueueBenchmark.test1Producer:test1Producer·p0.99                100  sample           19430.400            ns/op
MpscRelaxedArrayQueueBenchmark.test2Producers                                   100  sample  282486    8875.034 ±  14.771  ns/op
MpscRelaxedArrayQueueBenchmark.test2Producers:test2Producers·p0.99              100  sample           23104.000            ns/op
MpscRelaxedArrayQueueBenchmark.test3Producers                                   100  sample  324322   11579.367 ±  10.287  ns/op
MpscRelaxedArrayQueueBenchmark.test3Producers:test3Producers·p0.99              100  sample           12704.000            ns/op
```
P.S: the LogBuffer is writing much more than the queues but it doesn't suffer any write barrier while clearing/offering the messages of the buffer: this makes this experiment far from being fair, but it is good enough to show if the benefits of using XADD are similar to those of the original algorithm.

franz1981:
I've written a long running test with multiple producers and sometimes it got stuck with only few producers capable to make progress and it happens more often when a small (~128) capacity is configured: I'm trying to figure it out what's happening (probably a math error on my side while offering).

franz1981:
I've fixed  (and already squashed) the math error: the original heuristic built to allow long lived queues with small capacity wasn't good enough to cope with the concurrent backoff/overclaiming during a cycle rotation: I've changed it with another one that seems able to increase the queue lifecycle without exposing it to overclaims that could change the claimed cycleId.
If makes sense I could add a check on the maximum claimed position like this:
https://github.com/real-logic/aeron/blob/5ead930a6bd145acaf235a2905fbcbac19c89bae/aeron-client/src/main/java/io/aeron/Publication.java#L587
But throwing an exception when it happens.

franz1981:
I'm going to fix the claim initialisation on cycle rotation due to this: https://github.com/real-logic/aeron/pull/403

franz1981:
@mjpt777 I've just added a fix to address the same issue referenced on https://github.com/real-logic/aeron/commit/d55b9b8c78d95ba0d3c73398fb0a658699629261
Now the tests with multiple producer threads (128 on a 4 cores machine) are not failing and corrupting not consumed data anymore.

franz1981:
There is another issue related to this https://github.com/real-logic/aeron/commit/8735ef4609a26fe539b2ecb2118bd1e474383e6d that is needed to be ported too

franz1981:
I've added a commit to address it in a way similar to those on Aeron's LogBuffer.

franz1981:
I've added a commit to detect exhaustion of:
- position on cycle: mainly due to overclaiming (ie small capacity, big number of concurrent producers)
- claim id: mainly time related (ie small capacity and high throughput long running producers)

nitsanw:
Many thanks @franz1981 I've got some style tweaks and bits, but overall looks good

PR #201, opened 2017-11-12
 Fixed producer's claim on a blocked relaxed queue

franz1981:
I've provided:
- a fix on slow producer claims
- an improvement on rotation to detect (and warn) "bad" performers without blocking the q

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpscRelaxedArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscRelaxedArrayQueue.java
@@ -207,6 +207,11 @@ abstract class MpscRelaxedArrayQueueProducerCycleClaimFields<E> extends MpscRela
     {
         return UNSAFE.getAndAddLong(this, calcProducerCycleClaimOffset(cycleIndex), 1);
     }
+
+    protected final boolean casProducerCycleClaim(int cycleIndex, long expectedValue, long newValue)
+    {
+        return UNSAFE.compareAndSwapLong(this, calcProducerCycleClaimOffset(cycleIndex), expectedValue, newValue);
+    }
 }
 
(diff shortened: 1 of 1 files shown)


franz1981:
It includes the commit from https://github.com/JCTools/JCTools/pull/198 too: I will remove it if it will be merged or this one pass the review.

@nitsanw I need very much a review on that because this is pretty tough and I'm not sure I've found a reliable solution even if the old failing benchmarks seem fixed. :+1:

franz1981:
This [commit](https://github.com/JCTools/JCTools/pull/201/commits/81bbff455328bf26c418aa4ed8bdbcd303e0fab2) add an even more backpressure policy that could help with smaller capacity and big number of producers to not fail.

franz1981:
In the last version I've tried to mitigate the slow producers issue by employing 2 different strategies:

1. providing additional "relaxed" (aka best effort) backpressure vs any producers that offer beyond the allowed limits (consumerPosition + cycleLength)
2. allow the slow producers unable to rollback their claims to produce if position < consumerPosition + (cycleLength*2) (ie the real space left on the buffer)

I hope it is enough to make it more reliable: weird scheduling accidents and/or thread starvations are not simple to be managed and this one can't be considered a definitive solution but it allows the q to be used with tons of producers and a capacity of 2 for several bursts of messages (~ 1_000_000 on my laptop) without blocking the q (provided fast enough consumers).

nitsanw:
I'm merging this, but will shortly move the queue to experimental as I don't think it is ready for release.

franz1981:
Fair enough: the fix makes it more resilient to scheduling accident but it is not 100% safe like the others q :(
Re the fix will you take a look at it with me in the future? It is just for academic purposes but it's AFAIK the only porting of the Aeron LogBuffer...it deserves more love :)

PR #202, opened 2017-11-16
ConcurrentMap contract requires to give a result based on equals, not on reference equality

henri-tremblay:
According to the Javadoc, true should be returned in case of equality for both `remove` and `replace`.

```
if (map.containsKey(key) && Objects.equals(map.get(key), oldValue)) {
  map.put(key, newValue);
  return true;
} else
  return false;
```

 The replacement and removal are correctly done using `equals`, however the return value is currently based on reference equality. This patch solves this issue using `Objects.equals`.


nitsanw:
Thanks!

Issue #204, opened 2017-11-27
Document performance characteristics of SpscChunked/Growable on wiki


brettwooldridge:
Thanks for the project.

While the [Which Queue Should I Use?](https://github.com/JCTools/JCTools/wiki/Which-Queue-Should-I-Use%3F) wiki page is useful.  There are still subtlties that are not adequately explained; and it would be awesome if they could be added to that page.

For example, what are the performance differences/trade-offs between ``SpscChunkedArrayQueue`` and ``SpscGrowableArrayQueue``?

nitsanw:
Can you please provide a list of your question to more concretely direct the "subtlties that are not adequately explained"?
Thanks

brettwooldridge:
@nitsanw Clearly there is a difference between ``SpscChunkedArrayQueue`` and ``SpscGrowableArrayQueue``, otherwise two separate classes wouldn't exist.  However, it is not at all clear what those differences are.  Does one have higher throughput?  Does one use less memory?  Etc.

Additional explanation/documentation would be useful in helping users choose which is best for a given situation.

nitsanw:
Between the update to the wiki and the improved javadoc I think this is resolved.

brettwooldridge:
@nitsanw Thanks.

PR #206, opened 2018-01-19
Improved JavaDoc of linked array queues by adding more informations about relevant behaviours

franz1981:
I've added few bit to the original JavaDoc + missing ones to help issues like: https://github.com/JCTools/JCTools/issues/204


nitsanw:
Thanks for the contribution!

PR #211, opened 2018-05-25
Implement batching methods on MpmcArrayQueue

franz1981:
It adds batch drain/fill methods for MpmcArrayQueue

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java
@@ -128,10 +128,13 @@ abstract class MpmcArrayQueueL3Pad<E> extends MpmcArrayQueueConsumerIndexField<E
  */
 public class MpmcArrayQueue<E> extends MpmcArrayQueueL3Pad<E>
 {
+    public static final int MAX_LOOK_AHEAD_STEP = Integer.getInteger("jctools.mpmc.max.lookahead.step", 4096);
+    private final int lookAheadStep;
 
     public MpmcArrayQueue(final int capacity)
     {
         super(RangeUtil.checkGreaterThanOrEqual(capacity, 2, "capacity"));
+        lookAheadStep = Math.max(2, Math.min(capacity() / 4, MAX_LOOK_AHEAD_STEP));
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


franz1981:
@nitsanw I will add the batch fill as a separate commit with/without its own PR :+1:

jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java
```diff
@@ -128,10 +128,13 @@ final boolean casConsumerIndex(long expect, long newValue)
  */
 public class MpmcArrayQueue<E> extends MpmcArrayQueueL3Pad<E>
 {
+    public static final int MAX_LOOK_AHEAD_STEP = Integer.getInteger("jctools.mpmc.max.lookahead.step", 4096);
```
nitsanw: Should probably use `PortableJvmInfo.RECOMENDED_OFFER_BATCH` instead. On SPSC there's no reason not to claim large chunks for an uncontended producer, but for MPSC/MPMC the assumption should be some contention. Given that the empty slots in the queue will result in consumer 'bubbles' (queue not empty but slot is empty) we can potentially have an issue if a producer takes a while to fill up the slots it claims.

felixbarny:
Looking forward to this. When is the next release scheduled?

nitsanw:
Sorry for the slow response. Totally swamped ATM, will release around end of October.

franz1981:
@felixbarny I've implemented the batch methods assuming mostly single threaded usage to get the best of it: at the first failed attempt (due to contention) to fill/drain in batches it will just fill/drain one by one. 
it is what you're expecting?

felixbarny:
I'm using an MPMC queue as the basis for an object pool. My scenario is that I have multiple consumers and in 99% of the cases a single producer. In rare cases, there can be multiple producers.

I was hoping that both the producer and the consumers would benefit from batch operations. The producer could batch up multiple objects and add them to the object pool/queue in batch. Also, the consumers could get multiple objects and buffer them thread-local. By doing this, my expectation would be to gain performance by reducing the amount of CAS operations (1 per batch vs 1 per element).

franz1981:
> My scenario is that I have multiple consumers and, under normal circumstances, a single producer. In rare cases, there can be multiple producers.

I would say for sure that it will do it for the producers, given the "mostly single-threaded" scenario, will benefit a lot by batch fill 

> By doing this, my expectation would be to gain performance by reducing the amount of CAS operations (1 per batch vs 1 per element).

That's the ratio behind this PR indeed :) 
I just suggest to build up some benchmark/experiment with your scenarios to look if it is improving things as you expect and with different `jctools.mpmc.max.lookahead.step` parameters :+1:

PR #218, opened 2018-09-07
Fixes #216: Size of SpscGrowableArrayQueue can exceeds max capacity

### LINKED ISSUE #216: Size of queue may more then max size +1
I found a bug earlier this year

Seems to be

spscArrayGrowableQueue  

init give the max size 

but can run code make the queue = max +1

i truly run test code find it ,it is existence  
in Special cases

franz1981:
Adjust producer limit in order to allow one more element to be added
in place of the unneeded JUMP when the max capacity producer buffer
has been allocated


franz1981:
@nitsanw I can add a test similar to the one provided by @yangyi438 if necessary, but I have already verified that it fixes the referenced issue :)
Sorry for the multiple changes of commit msg/PR name, but I was experimenting to reference the issue number into the commit msg to make allow github to ref & close them in tandem :P

nitsanw:
Please add the test.
Thanks for jumping in :-)

franz1981:
@nitsanw No worries :)
I admit is not one of my most brilliant tests, but let me know if is enough :+1:

PR #221, opened 2018-10-10
Fixes #220: MpscCompound tests must account for CPUs not power of 2

### LINKED ISSUE #220: MpqSanityTestMpscCompound>MpqSanityTest.testSizeIsTheNumberOfOffers:190 expected:<72> but was:<128>
Happens on an actual computer with 2x18 core with SMT.
Somehow, this test assumes that `spec.capacity` (set to `Runtime.getRuntime().availableProcessors()` in `PortableJvmInfo`) conforms to the number of relaxedOffers that can be made until `SIZE` if reached or `relaxedOffer `is denied. 
In `relaxedOffer`, we get to make two offers in each queue where for each queue, the switchover to the next queue occurs after the second offer. 
There are 64 queues created with a capacity of 2 because of some `roundToPowerOfTwo `voodoo in `MpscCompoundQueueColdFields(int capacity, int queueParallelism)`. Of course, these add up to 128 which is not 72, hence the fail.
This should be reproducible on any machine with a non-power-of-two number of hardware threads.
Not sure if this particular test is representative for any important assumptions or if it can be ignored for all practical purposes.

franz1981:


nitsanw:
Thanks!

PR #222, opened 2018-10-19
Fix typo and gitignore

nastra:


nastra:
ping @nitsanw in case you have a sec for a review

nitsanw:
Thanks! sorry for the slow reply

PR #223, opened 2018-11-06
Better way to get shift factor

mburger81:


nitsanw:
This is better in some sense of better, but it is a bit less readable IMO. I would be open to accepting this with a nice set of comments explaining what are expected values and why this works.

davidmoten:
Does this provide a performance benefit?

davidmoten:
Excuse my curiosity, I'm not directly affected by this change but I'm confused as to why it is better.

franz1981:
I believe that for "better" it means that's more generic, but it is unlikely to happen that the scale will be different from that numbers (4, 8) :)

nitsanw:
@mburger81 ?

Issue #228, opened 2019-01-17
Add method to Mpsc(Unbounded)ArrayQueue to allow non-consumer thread to get a snapshot of the queue content


srdo:
Hi,

I would like to be able to peek at the contents of the MpscArrayQueue or MpscUnboundedArrayQueue from a non-consumer thread.

I'm working on Apache Storm, which has the following queue setup to process a stream of data:

Inbound thread -> inputQueue -> user code -> outputQueue -> Outbound thread

I would like for a thread that is not the user code or the outbound thread to be able to peek at, but not modify, the elements of inputQueue and outputQueue. Essentially I would like a `public Collection<E> unorderedSnapshot()` method that just returns a copy of the contents of the queue's buffer, filtering out any nulls or SKIP instructions.

Just to justify why I need this, feel free to skip if it's not important:

Storm is a system for distributed processing of data streams. When a new element in a data stream is submitted for processing, it may go to many physical machines before it is considered fully processed. 

Storm puts a timeout on how long the full processing sequence is allowed to take, in order to offer at-least-once processing guarantees in the presence of e.g. crashing machines or network issues.

I would like to be able to keep this timeout from expiring as long as any machines are still actively processing an element. One of the places an element is at risk of timing out is when it is stuck in the inputQueue or outputQueue, either due to slow user code, or due to backpressure from downstream processing steps.

To keep the timeout from expiring, I would like for a heartbeat thread to occasionally reset the timeout for messages that are stuck in the inputQueue or outputQueue. In order to do this, I need this thread to be able to read the queue contents.

nitsanw:
This could be implemented if the queues supported iterator in a `weak` fashion. The reason iterators were not implemented for JCTools queues are:
 1. Implementing `remove` is terrible array backed queues 
 2. Extra work for not much interest
 3. Inconsistencies with concurrent access

If we agree on not implementing `remove` and offering no guarantees with regards to concurrent access I think this is fairly straight forwards. The weak iterator would be only work properly on queues at rest (i.e. that are not being modified concurrently), but it would be hard to enforce no early termination in the face of consumers. Consider the case where a linked node is consumed in parallel to iteration: If we do not clear the linking reference we will cause GC nepotism, but if we do clear it we cause iterators to abort mid queue. Alternatively consider iteration on an array backed buffer where consumers/producer may work their way around the buffer before the iterator is moved. One thing that is not acceptable to my mind is hampering the main usecase(message passing) to support iterators.

[...]

PR #229, opened 2019-01-17
Issue-228: Add unorderedSnapshot method to MpscArrayQueue and BaseMps…

srdo:
…cLinkedArrayQueue

Resolves https://github.com/JCTools/JCTools/issues/228


nitsanw:
See comment on main ticket on API.

jctools-core/src/main/java/org/jctools/queues/BaseMpscLinkedArrayQueue.java
```diff
@@ -403,7 +417,6 @@ else if (casProducerIndex(pIndex, pIndex + 1))
     {
         final long offset = nextArrayOffset(mask);
         final E[] nextBuffer = (E[]) lvElement(buffer, offset);
-        soElement(buffer, offset, null);
```
nitsanw: This will cause GC nepotism

jctools-core/src/main/java/org/jctools/queues/BaseMpscLinkedArrayQueue.java
```diff
@@ -73,13 +77,22 @@ final boolean casProducerIndex(long expect, long newValue)
     private volatile long consumerIndex;
     protected long consumerMask;
     protected E[] consumerBuffer;
+    private volatile E[] volatileConsumerBuffer;
```
nitsanw: why do we need 2 consumer buffers?

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
```diff
@@ -562,4 +565,17 @@ public void fill(Supplier<E> s, WaitStrategy w, ExitCondition exit)
             idleCounter = 0;
         }
     }
+    
+    public List<E> unorderedSnapshot() {
+        int length = capacity();
+        List<E> elements = new ArrayList<E>();
+        for (int i = 0; i < length; i++) {
```
nitsanw: You iterate through the whole array, very likely to be many thousands of elements, to find the elements you need

nitsanw:
If we're going to add this access we will need some very clearly worded java doc on it's limitations and guarantees.

jctools-core/src/main/java/org/jctools/queues/BaseMpscLinkedArrayQueue.java
srdo: Should I make consumerBuffer volatile and add lp/lv/svConsumerBuffer methods instead?

jctools-core/src/main/java/org/jctools/queues/BaseMpscLinkedArrayQueue.java
srdo: Thanks, wasn't aware of this issue. Will try to come up with a solution

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
srdo: I'll try to make it less wasteful.

jctools-core/src/main/java/org/jctools/queues/BaseMpscLinkedArrayQueue.java
nitsanw: why does it need to be volatile?

srdo: I wasn't aware that non-volatile reference writes/reads were always atomic. I guess it doesn't need to be volatile. Thanks.

jctools-core/src/main/java/org/jctools/queues/BaseMpscLinkedArrayQueue.java
srdo: I replaced this with a marker object that will cause the iterator to jump to the current consumerBuffer.

srdo:
I've updated to provide iterators instead, and believe the GC nepotism issue should be avoided.

I made a minor change to the order of writes when replacing the consumer buffer. I would like to write the consumerBuffer field before the BUFFER_CONSUMED marker is inserted, to ensure the iterator doesn't see the BUFFER_CONSUMED flag along with the old consumerBuffer value, which would cause it to scan the same buffer again. Let me know if this isn't okay, and I'll be happy to revert.

I couldn't figure out a way to avoid scanning the entire buffer in the linked queues. I'd like to avoid skipping elements that are still queued, and as far as I can tell we will risk doing so if we follow the consumer index. If the iterator sees a mismatch between the consumerIndex and the current buffer, e.g. because the consumer just replaced consumerBuffer, it may start at the wrong position and end up skipping elements.

nitsanw:
Sorry for taking long to re-review, this looks good :-)

PR #230, opened 2019-01-18
XADD Chunked Array Queue

franz1981:
It is a chunked array q designed to scale with the number of producers.

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpscProgressiveChunkedQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscProgressiveChunkedQueue.java
@@ -1,0 +1,704 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jctools.queues;
+
+import org.jctools.queues.IndexedQueueSizeUtil.IndexedQueue;
+import org.jctools.util.PortableJvmInfo;
+import org.jctools.util.Pow2;
+import org.jctools.util.UnsafeRefArrayAccess;
+
+import java.util.AbstractQueue;
+import java.util.Iterator;
+
+import static org.jctools.util.UnsafeAccess.UNSAFE;
+import static org.jctools.util.UnsafeAccess.fieldOffset;
+
+
+abstract class MpscProgressiveChunkedQueuePad1<E> extends AbstractQueue<E> implements IndexedQueue
+{
+    long p01, p02, p03, p04, p05, p06, p07;
+    long p10, p11, p12, p13, p14, p15, p16, p17;
+
+    static final class AtomicChunk<E>
+    {
+        final static int NIL_CHUNK_INDEX = -1;
+        private static final long PREV_OFFSET = fieldOffset(AtomicChunk.class, "prev");
+        private static final long NEXT_OFFSET = fieldOffset(AtomicChunk.class, "next");
+        private static final long INDEX_OFFSET = fieldOffset(AtomicChunk.class, "index");
+        private volatile AtomicChunk<E> prev;
+        private volatile long index;
+        private volatile AtomicChunk<E> next;
+        private final E[] buffer;
+
+        AtomicChunk(long index, AtomicChunk<E> prev, int size)
+        {
+            buffer = CircularArrayOffsetCalculator.allocate(size);
+            spNext(null);
+            spPrev(prev);
+            spIndex(index);
+        }
+
+        final AtomicChunk<E> lvNext()
+        {
+            return next;
+        }
+
+        final AtomicChunk<E> lpPrev()
+        {
+            return (AtomicChunk<E>) UNSAFE.getObject(this, PREV_OFFSET);
+        }
+
+        final long lvIndex()
+        {
+            return index;
+        }
+
+        final void soIndex(long index)
+        {
+            UNSAFE.putOrderedLong(this, INDEX_OFFSET, index);
+        }
+
+        final void spIndex(long index)
+        {
+            UNSAFE.putLong(this, INDEX_OFFSET, index);
+        }
+
+        final void soNext(AtomicChunk<E> value)
+        {
+            UNSAFE.putOrderedObject(this, NEXT_OFFSET, value);
+        }
+
+        final void spNext(AtomicChunk<E> value)
+        {
+            UNSAFE.putObject(this, NEXT_OFFSET, value);
+        }
+
+        final void spPrev(AtomicChunk<E> value)
+        {
+            UNSAFE.putObject(this, PREV_OFFSET, value);
+        }
+
+        final void soElement(int index, E e)
+        {
+            UnsafeRefArrayAccess.soElement(buffer, UnsafeRefArrayAccess.calcElementOffset(index), e);
+        }
+
+        final void spElement(int index, E e)
+        {
+            UnsafeRefArrayAccess.spElement(buffer, UnsafeRefArrayAccess.calcElementOffset(index), e);
+        }
+
+        final E lvElement(int index)
+        {
+            return UnsafeRefArrayAccess.lvElement(buffer, UnsafeRefArrayAccess.calcElementOffset(index));
+        }
+
+    }
+}
+
+// $gen:ordered-fields
+abstract class MpscProgressiveChunkedQueueProducerFields<E> extends MpscProgressiveChunkedQueuePad1<E>
+{
+    private final static long P_INDEX_OFFSET =
+        fieldOffset(MpscProgressiveChunkedQueueProducerFields.class, "producerIndex");
+    private volatile long producerIndex;
+
+    @Override
+    public final long lvProducerIndex()
+    {
+        return producerIndex;
+    }
+
+    final long getAndIncrementProducerIndex()
+    {
+        return UNSAFE.getAndAddLong(this, P_INDEX_OFFSET, 1);
+    }
+
+    final long getAndAddProducerIndex(long delta)
+    {
+        return UNSAFE.getAndAddLong(this, P_INDEX_OFFSET, delta);
+    }
+}
+
+abstract class MpscProgressiveChunkedQueuePad2<E> extends MpscProgressiveChunkedQueueProducerFields<E>
+{
+    long p01, p02, p03, p04, p05, p06, p07, p08;
+    long p10, p11, p12, p13, p14, p15, p16;
+}
+
+// $gen:ordered-fields
+abstract class MpscProgressiveChunkedQueueProducerBuffer<E> extends MpscProgressiveChunkedQueuePad2<E>
+{
+    private static final long P_BUFFER_OFFSET =
+        fieldOffset(MpscProgressiveChunkedQueueProducerBuffer.class, "producerBuffer");
+    private static final long P_CHUNK_INDEX_OFFSET =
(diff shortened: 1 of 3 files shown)


franz1981:
@nitsanw It is a just a lil' more then a POC but I will continue to improve it if the logic seems reasonable and sound to you, so please take a look to the logic :+1: 
I have yet to fix:

1. tests
2. implements missing methods
3. use standard naming
4. use Unsafe where necessary
5. padding to protect from false sharing
6. "maybe" Implement AtomicChunk::next using a JUMP + AtomicChunk element

The first benchmarks shows a better scaling then the other multi-producer queues and the code/logic/layout is not optimized at all :+1:

franz1981:
I will probably reuse some of the already existing tests, but as I have written above is just a pr to get a review on the q logic; I will squash a more refined version I'd the logic will be validated...

franz1981:
I'm working on an idea to recycle the consumed buffers and make it GC-free in the same way MpscChunkedArrayQueue is :+1:

franz1981:
@eolivelli @nitsanw I've played a lil' more with this on the weekend, making possible to recycle chunks, adding tests and paggings + refine the implementation in order to use Unsafe.
The performances are interesting: with 2 producers I'm getting 20 ops/us vs 13 ops/usec of MpscUnboundedArrayQueue + no garbage produced.
With 7 producers I'm getting 40 ops/us vs 8 ops/usec of MpscUnboundedArrayQueue + near no garbage produced (depends on long stall of any producer thread).

franz1981:
I think that both poll/peek and fill could be improved a lot; single threaded fill benchs are not good as I was expecting!

jctools-core/src/main/java/org/jctools/queues/MpscProgressiveChunkedQueue.java
```diff
@@ -0,0 +1,553 @@
...
+    @Override
+    public E poll()
+    {
+        final long chunkMask = this.chunkMask;
+        final long consumerIndex = this.lpConsumerIndex();
+        AtomicChunk<E> consumerBuffer = this.consumerBuffer;
+        final int consumerOffset = (int) (consumerIndex & chunkMask);
+        if (consumerOffset == 0 && consumerIndex >= (chunkMask + 1))
+        {
+            AtomicChunk<E> next = consumerBuffer.lvNext();
```
nitsanw: Cold path, move to separate method?

jctools-core/src/main/java/org/jctools/queues/MpscProgressiveChunkedQueue.java
```diff
@@ -0,0 +1,553 @@
...
+            //that a racing producer can already obtain it
+            if (lpFreeBuffer() == null)
+            {
+                soFreeBuffer(consumerBuffer);
+            }
+            next.spPrev(null);
+            consumerBuffer = next;
+            this.consumerBuffer = consumerBuffer;
+        }
+        E e = consumerBuffer.lvElement(consumerOffset);
```
nitsanw: The following code can be re-arranged as elsewhere for single consumer code

jctools-core/src/main/java/org/jctools/queues/MpscProgressiveChunkedQueue.java
```diff
@@ -0,0 +1,553 @@
...
+    public int drain(Consumer<E> c)
+    {
+        return drain(c, 4096);
+    }
+
+    @Override
+    public int fill(Supplier<E> s)
+    {
+        long result = 0;// result is a long because we want to have a safepoint check at regular intervals
+        final int capacity = 4096;
```
nitsanw: Why not default to chunk size here?

jctools-core/src/main/java/org/jctools/queues/MpscProgressiveChunkedQueue.java
```diff
@@ -0,0 +1,553 @@
...
+        consumerBuffer = first;
+        chunkMask = chunkSize - 1;
+        chunkShift = Integer.numberOfTrailingZeros(chunkSize);
+        soFreeBuffer(new AtomicChunk(-1, null, chunkSize));
+    }
+
+    private AtomicChunk<E> producerBufferOf(AtomicChunk<E> producerBuffer, long chunkIndex)
+    {
+        long jumpBackward = producerBuffer == null ? -1 : producerBuffer.index - chunkIndex;
+        while (jumpBackward < 0)
```
nitsanw: You can optimize for common case here where the index matches (jumpBackward == 0)

nitsanw:
Great stuff! This makes for an interesting tradeoff, though the progress is still blocked if the new buffer allocator is interrupted (so this is still blocking producers on new buffer). The progress guarantees are the same overall, so this is not more "progressive" IMO. You can make it more progressive if the first to discover a missing chunk allocates it, rather than the first in chunk, but this is not simple...

franz1981:
> though the progress is still blocked if the new buffer allocator is interrupted (so this is still blocking producers on new buffer).

Eheh, smart observation: that's why I've added the recycling of the buffer; to speedup the process to publish a new chunk to be used by the other producers awaiting it

> The progress guarantees are the same overall, so this is not more "progressive" IMO

You're right, but I wasn't able to find a better name for this :P
MostlyProgressive seems even worst eheh

> You can make it more progressive if the first to discover a missing chunk allocates it, rather than the first in chunk, but this is not simple...

Nope, but it is an interesting idea! Not sure I will be able to get it into this PR

jctools-core/src/main/java/org/jctools/queues/MpscProgressiveChunkedQueue.java
franz1981: So, perform the same check outside this slow path method to avoid it to be called?

jctools-core/src/main/java/org/jctools/queues/MpscProgressiveChunkedQueue.java
franz1981: Do you mean by grouping everything under a single method or just by grouping the "slow" path:
```java
        assert !firstElementOfNewChunk;
        if (lvProducerIndex() == consumerIndex)
        {
            return null;
        }
        e = spinForElement(consumerBuffer, consumerOffset);
        consumerBuffer.spElement(consumerOffset, null);
        soConsumerIndex(consumerIndex + 1);
        return e;
```

franz1981:
@nitsanw I've added some significant changes to the producerBuffer publication/initialization logic due to the complication of of the recycling of old instances, but I think to have addresses any weird corner cases due to long stalls of producer threads. Please let me know if the comments are clear enough: I wasn't expecting recycling logic to affect the logic that much!

franz1981:
I'm just too curious to see how it performs on a machine with many cores: @nitsanw there is any chance to test this one on such hw?

franz1981:
@nitsanw I've addressed in a separate commit the challenge:
> though the progress is still blocked if the new buffer allocator is interrupted (so this is still blocking producers on new buffer). The progress guarantees are the same overall, so this is not more "progressive" IMO. You can make it more progressive if the first to discover a missing chunk allocates it, rather than the first in chunk, but this is not simple...

Now it should be more progressive: any offer that found any missing chunk (including previous ones) will try to fill the gap, but given that producer buffer publication (partial) ordering should be preserved I've used a CAS on a separate `producerChunkIndex` to maintain ordering of publish.
Maybe there is a better way to make it totally progressive, but I will probably need a lot more time to find it :)

jctools-core/src/main/java/org/jctools/queues/MpscProgressiveChunkedQueue.java
```diff
@@ -0,0 +1,704 @@
...
+        assert producerBuffer.lvIndex() == expectedChunkIndex;
+        return producerBuffer;
+    }
+
+    private AtomicChunk<E> appendNextChunk(AtomicChunk<E> producerBuffer, long chunkIndex, int chunkSize)
+    {
+        assert chunkIndex != AtomicChunk.NIL_CHUNK_INDEX;
+        final long nextChunkIndex = chunkIndex + 1;
+        //prevent other concurrent attempts on appendNextChunk
+        if (!casProducerChunkIndex(chunkIndex, nextChunkIndex))
```
franz1981: If we want a much more robust code here it can be replaced with 
`casProducerChunkIndex(chunkIndex, AtomicChunk.NIL_CHUNK_INDEX)` and a 
`soProducerChunkIndex(nextChunkIndex)` at the end IF no OOM error will happen :)
If a OOM error will happen it would re-set the original value `soProducerChunkIndex(nextChunkIndex)`,
leaving other threads able to re-try to allocate a new chunk or (if lucky enough) that the consumer
will provide a recycled chunk in time :)
I see that the other linked queues are not providing protection vs OOM, that's not strictly necessary, but 
given that are using `cas` it shouldn't complicate that much the impl

franz1981:
@eolivelli @felixbarny did you have chances to try it or check the logic guys?

felixbarny:
Sorry, but I didn't try it out yet. For my use case, it's important that the size of the queue is limited. This one seems to be unlimited? Ideally, I'd also need a XADD-based MPMC queue.

`<offtopic>`
A quick outline of my use case. I'm writing an APM agent. It needs to collect traces from multiple application threads which are transferred to a reporter thread. After reporting, the objects are reset and put in an object pool for reuse. When a request is recorded, the agent takes a recycled object from the pool in the context of an application thread. I'm currently using a MPSC disruptor to transfer the objects from the application threads to the reporter thread. For the object pool, I'm using a MPMCArrayQueue. This design is quite simple but getting objects from the pool and putting them into the disruptor is contended. I have some ideas to reduce the contention with ThreadLocals but that would be even more off-topic here. A XADD-based MPMC queue would allow the architecture to remain simple but still not be prone to contention.
`</offtopic>`

franz1981:
I have already implemented a bounded mpsc xadd queue but is prone to be blocked forever in very special cases ie is in the experimental branch of JCTools foe thwt reason and is insoired on the Aeron log buffer algorithm.
I can't ATM see a chance to preserve a queue semantic AND using XADD on producer side: without being too negative I think they is difficult to achieve it with a bounded q, that's why I have built this one.

This q could be "easily" turn into mpmc (but without using XADD on consumer side obviously) and I could provide a weakOffer method that could just weakly back-pressure a producer by looking the current distance of consumers, reducing the chance to allocate a new chunks.
I'm not sure it is the right tool TBH while using a mpmc bounded queue with batch drain/fill should work a lot better.

franz1981:
@kay I have alway referenced Nitsan for the review of this PR, but please, feel free to validate the code as well eh :+1:

franz1981:
@nitsanw I've pushed a squashed commit addressing a null check, but the logic should be the same.
The first commit, that wasn't performing cooperative/progressive allocation of missing chunks isn't needed anymore IMHO, but I'm lefting it here just while awaiting the code review: I will squash the commits into one right after that :+1:

franz1981:
I've decided to squash the commits into one, given that the "more progressive" version is preferreable to the original one,@nitsanw let  me know if I'm missing other parts to clean-up or any improvement you see on the algorithm/implementation.
I see that the `freeBuffer` could be replaced by a `SpscArrayQueue` configurable at startup and many volatile loads could be replaced, but my only concerns perf-wise are the batch operations that doesn't seem to me effective as I was expecting.

franz1981:
@nitsanw I've tried a scaling benchmark with increasing numbers of producers, but I'm not getting what I was expecting:
image
According to http://www.perfdynamics.com/Manifesto/USLscalability.html is quite clear that there are some other costs in place that are reducing the scalability of this q.
The fist part of the shape of the chart is perfectly matching the classic USL (for some coherency and contention factors to be estimated):
image
But what seems the most intriguing part to me is that I've found an evident pattern for the tail of the chart
and it seems somehow related to [Gene Amdahl](https://en.wikipedia.org/wiki/Gene_Amdahl) himself:
image
After several hours I can show you my bet:
image
It is clear that the Universal Scalability Law is clearly an extension of Amdahl Law, but Amdhal himself is a personified extension of it as well (!!!).
I'm sure that Little himself could be hidden somewhere in my bechmark results...
maybe next 1st April I'll find out :)

nitsanw:
I'll review further before next release, but generally it looks good :-)
Thanks for your efforts and patience

PR #231, opened 2019-03-01
Fixes #208: MpscLinkedAtomicQueue can be made not final

### LINKED ISSUE #208: Reason why the MpscLinkedAtomicQueue class is final?
I'm implementing time sliced self processing queues, and this is the only instance of the backing queue for which I need to wrap instead of extending the class. I'm fine with methods being final, but I cannot understand why the class itself cannot be extended in this specific case?

franz1981:


franz1981:
@nitsanw 72c55e5bee5ed05e47c5f997f8d4bc9d736659d1 has reintroduced `final`

PR #235, opened 2019-03-03
Polish - use try-with-resources and simplify name calculation

JanStureNielsen:


kay:
Could be done on `JavaParsingAtomicLinkedQueueGenerator` too

PR #241, opened 2019-04-03
Protect the producer index in case of OutOfMemoryError

franz1981:


franz1981:
I would like to add tests for this, but probably I need to mock the allocate function in order to do it

nitsanw:
LGTM, thanks

PR #245, opened 2019-05-22
Fixes #226: XADD mpmc queue: highly scalable queue

### LINKED ISSUE #226: Queues based on Unsafe#getAndAdd instead of CAS spins?
Hi,

I'm learning about lock-free and wait-free algorithms and I was asking myself whether queues could be implemented based on the wait-free `Unsafe#getAndAdd` method as opposed to the lock-free CAS spin. The latter can still lead to queuing effects, especially under high contention and when there are more threads than processors concurrently doing the CAS spin.

Example of a CAS spin in `MpmcArrayQueue`:
https://github.com/JCTools/JCTools/blob/6b8feeaa913c780fd560943cf2d0d8f0af8461c2/jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java#L175-L176

Would it be possible to restructure the algorithm to be wait-free or is that inherently impossible?

Thanks,
Felix

franz1981:
It is based on a similar algorithm of the mpsc variant, but with a simple thread-safe polling mechanism (not progressive in any of its parts ATM).

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpmcProgressiveChunkedQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpmcProgressiveChunkedQueue.java
@@ -1,0 +1,781 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jctools.queues;
+
+import org.jctools.queues.IndexedQueueSizeUtil.IndexedQueue;
+import org.jctools.util.PortableJvmInfo;
+import org.jctools.util.Pow2;
+import org.jctools.util.UnsafeAccess;
+import org.jctools.util.UnsafeRefArrayAccess;
+
+import java.util.AbstractQueue;
+import java.util.Arrays;
+import java.util.Iterator;
+
+import static org.jctools.util.UnsafeAccess.UNSAFE;
+import static org.jctools.util.UnsafeAccess.fieldOffset;
+
+
+abstract class MpmcProgressiveChunkedQueuePad1<E> extends AbstractQueue<E> implements IndexedQueue
+{
+    long p01, p02, p03, p04, p05, p06, p07;
+    long p10, p11, p12, p13, p14, p15, p16, p17;
+
+    static final class AtomicChunk<E>
+    {
+        private static final long ARRAY_BASE;
+        private static final int ELEMENT_SHIFT;
+
+        static
+        {
+            final int scale = UnsafeAccess.UNSAFE.arrayIndexScale(long[].class);
+            if (8 == scale)
+            {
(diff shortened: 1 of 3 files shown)


franz1981:
@nitsanw When the polling algorithm will be sound enough (waiting your reciew) I will refactor it in order to:
- share code with mpsc 
- cleanup name variables in order to make simpler to automatic produce the Atomic variant 
- "maybe" make freeBuffer a SpscArrayQueue (or a simplificated version of it)

franz1981:
There is something I need to fix: I'm hitting
```OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
Exception in thread "Thread-2" java.lang.AssertionError
	at org.jctools.queues.MpmcProgressiveChunkedQueue.producerBufferOf(MpmcProgressiveChunkedQueue.java:310)
	at org.jctools.queues.MpmcProgressiveChunkedQueue.offer(MpmcProgressiveChunkedQueue.java:374)
	at org.jctools.queues.MpmcProgressiveChunkedQueue.relaxedOffer(MpmcProgressiveChunkedQueue.java:596)
	at org.jctools.queues.MpqSanityTest$9.run(MpqSanityTest.java:586)
	at java.lang.Thread.run(Thread.java:748)```
on org.jctools.queues.MpqSanityTestMpmcProgressiveChunked::testRelaxedOfferPollObservedSize

franz1981:
The issue was due to buffer recycling, causing new consumers to consume items before new ones will null out them. The fix is not elegant: a chunk could be reused only if all consumers has finished (ie nulling out elements), hence I've introduced old school ref counting :(
I have a strong feeling that using a long sequence into `AtomicChunk` (as MpmcArrayQueue) to 
force coordination among consumers/producers at slot level would be more elegant, although increasing memory footprint.

franz1981:
I've changed the ref counting implementation introducing several "improvements":
- pooled chunks that won't impose `new long[chunkSize]` additional memory footprint to all chunks (unpooled too)
- freeBuffer with configurable capacity thanks to SpscArrayQueue
- distributed sequencing per-slot to coordinate old consumers and new producers working on the same pooled chunk (simpler then MpmcArrayQueue)

franz1981:
Looking at the code with more attention I see that the consumer side could be made more progressive, by letting any consumer able to to cas the consumer buffer to the next one to reduce the distance from the expected chunkIndex (similarly to what we do on producer side): ATM I don't have much time to implement it :(, but I'm leaving this note for anyone that is interesting into this improvement (or myself of the future)

franz1981:
@nitsanw Did you have already taken a look to the algorithm bud? If not, do you want me to add some more informations on the JavaDoc about the logic?

nitsanw:
I'm waiting for you to finish

franz1981:
> I'm waiting for you to finish

I think I will stop for now to change it, I promise: too many things going on during my daily job :)
If you just wait me to finish, there is an high risk I will going into the infinite loop of: 
..-> optimize -> break tests -> deptimize&change idea -> revert- ->...
:P
Let me know if there is anything I can help with :+1:

nitsanw:
I'm going to put it in experimental and park it there

franz1981:
@nitsanw 
> I'm going to put it in experimental and park it there

Any concern about the algorithm? It shouldn't have any of the issues of the one inspired by the Aeron log buffer, but I could be wrong...

nitsanw:
@franz1981 I haven't had time to look at it closely enough, but now that I looked at the benchmark results... WOW :-)
This is a very interesting result, and definitely worth a closer look. Will spend some time on this in the next couple of days. I'd like to include the 2 new queues in the next release, which I'd like to push in the next couple of weeks.
Great work mate!

Issue #244, opened 2019-05-18
AssertionError when replacing missing key


fvlad:
Hi. I'm using jctools 2.1.2.
I found that when I try to replace a value in NBHM for a key that doesn't exit, I get AssertionError. Here is the sample test:

```
@Test
public void replaceTest() {
    final ConcurrentMap<String, String> map = new NonBlockingHashMap<>();

    assertNull(map.replace(UUID.randomUUID().toString(), UUID.randomUUID().toString()));
    assertFalse(map.replace(UUID.randomUUID().toString(), UUID.randomUUID().toString(), UUID.randomUUID().toString()));
}
```

The test works fine when I replace NBHM with ConcurrentHashMap.

PR #246, opened 2019-05-28
Fix replace missing value

fvlad:
Fix to the issue https://github.com/JCTools/JCTools/issues/244 to avoid throwing AssertionError when value is missing by key on replace

### DIFF
--- a/jctools-core/src/main/java/org/jctools/maps/NonBlockingHashMap.java
+++ b/jctools-core/src/main/java/org/jctools/maps/NonBlockingHashMap.java
@@ -379,7 +379,6 @@ public class NonBlockingHashMap<TypeK, TypeV>
     if (oldVal == null || newVal == null) throw new NullPointerException();
     final Object res = putIfMatch( this, _kvs, key, newVal, oldVal );
     assert !(res instanceof Prime);
-    assert res != null;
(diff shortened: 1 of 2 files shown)


nitsanw:
Javadoc for replace:
"replace(K key, V value): Replaces the entry for the specified key only if it is currently mapped to some value."
So, we expect a null reply, and the assert in putIfMatch seems to contradict the contract.

The code in `putIfMatch` has an assert that is stronger than the guarantees in the static `putIfMatch`, but that assert is valid for all case where the `oldVal` param is not `MATCH_ANY`. I think the right move here is to add this observation to the assert:
```
assert res != null || oldVal == MATCH_ANY;
```

nitsanw:
Further digging after https://github.com/JCTools/JCTools/pull/246#issuecomment-496482459 leads me to think that putIfMatch should have returned TOMBSTONE instead of null here.

nitsanw:
Please see changes in https://github.com/JCTools/JCTools/commit/1475be5e072677ecb657bd80649025a00fdf43bd which resolve the issue. I'm very grateful for the PR and would happily merge it if you revert the change to NBHM. Thanks!

fvlad:
Great, thanks for your fix!

PR #247, opened 2019-06-11
MpscBlockingConsumerArrayQueue interruption fixes

njhill:
After consumer is interrupted in `take()` method:
- Ensure queue reverts to original non-blocked state (revert producer index and blocking field) - otherwise for example a subsequent `poll()` will spin indefinitely
- Clear interrupt flag (consistent with equivalent j.u.c. behaviour)

Extended unit tests to cover this

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
@@ -364,17 +364,32 @@ public class MpscBlockingConsumerArrayQueue<E> extends MpscBlockingConsumerArray
             long pIndex = lvProducerIndex();
             if (cIndex == pIndex && casProducerIndex(pIndex, pIndex + 1))
             {
-                // producers only try a wakeup when both the index and the blocked thread are visible
-                soBlocked(Thread.currentThread());
-                do
+                boolean unblocked = false;
+                try
                 {
-                    LockSupport.park();
-                    if (Thread.currentThread().isInterrupted())
+                    // producers only try a wakeup when both the index and the blocked thread are visible
+                    soBlocked(Thread.currentThread());
+                    do
                     {
-                        throw new InterruptedException();
+                        LockSupport.park();
+                        if (Thread.interrupted())
+                        {
+                            throw new InterruptedException();
+                        }
+                    }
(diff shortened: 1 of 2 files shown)


njhill:
Thanks @franz1981, that only applies to the second "fix" though right? the first one does appear to be a clear bug to me at least.

I understand the goal isn't to impl `BlockingQueue` but would have thought it may as well mirror the semantics for the subset that overlaps? And clearing the interruption flag in this case isn't confined to blocking queue methods, it seems to be the general convention when `InterruptedException` is thrown.

franz1981:
> Thanks @franz1981, that only applies to the second "fix" though right? the first one does appear to be a clear bug to me at least.

Yep, sorry I haven't written clearly but yes, I'm +1 for the bug issue fix and personally to the whole PR :+1:

jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
```diff
@@ -364,17 +364,32 @@ public E take() throws InterruptedException
...
+                        {
+                            throw new InterruptedException();
+                        }
+                    }
+                    while (lvBlocked() != null);
+                    unblocked = true;
+                }
+                finally
+                {
+                    if (!unblocked) {
```
nitsanw: I think the diff would be cleaner/clearer if the boolean is removed and instead we incorporate the logic into the `Thread.interrupted` check. WDYT?

nitsanw:
Hi Nick, thanks for the contribution :-)
Please consider my style comment, if you decide to keep code as is I am happy to merge.

jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
njhill: Thanks @nitsanw I agree and actually that was my first pass, but the rationale for this approach is:
- It's more obviously robust, in that it's clear the queue is always left in a non-corrupt state without having to inspect and reason about what other obscure `Error`s may or may not be possible
- It nicely generalizes to the other `poll(long,TimeUnit)` PR impl, where there's a second "normal" abort condition, and seemed sensible for the logic to be consistent between the two

But I'm of course happy to go with whatever you prefer.

PS big fan of your work!

PR #248, opened 2019-06-11
Add MpscBlockingConsumerArrayQueue::poll(long, TimeUnit) method

njhill:
This is a minor variation of the existing `take()` impl. If the addition is of interest then I can add a unit test.

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
@@ -15,6 +15,7 @@ package org.jctools.queues;
 
 import java.util.AbstractQueue;
 import java.util.Iterator;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.LockSupport;
 
 import org.jctools.queues.IndexedQueueSizeUtil.IndexedQueue;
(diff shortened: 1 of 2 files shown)


njhill:
Thanks @franz1981 that's very interesting. My feeling is that the standard API would be beneficial regardless, and like you describe there are ways that timer resolution can be increased if/where needed, i.e. by separately setting a smaller slack value for the JVM. There is some tradeoff in terms of cpu consumption I guess. Supporting a custom park strategy also sounds like an interesting idea.

franz1981:
@njhill I haven't elaborated enough what I mean, sorry...
I mean that if we know that we are going to wait smal amount of time, or that a spurious wakeup on park will left us waiting a too small value, maybe we could avoid waiting and just spinning/yielding (not a great solution, but better then wait 50 us instead of 10 ns, probably) :)

nitsanw:
Looks good (style comment from prev. PR holds), please add tests. As for @franz1981 concerns, I hear ya, but this API is not the place to address those IMO. Yes we could for instance not bother sleeping for < 250ns or some other arbitrary value, or recommend people tune the timerslack, but the API suggests JUC behavior, so we should follow their lead. There's certainly room for an extra method with a "SleepStrategy" maybe, but lets leave it for another day.

njhill:
Thanks @nitsanw. I just pushed a minor update to refine the timeout arg handling, which makes it consistent with other j.u.c. impls e.g. failing fast for null time unit.

This coincidentally demonstrates my motivation behind the try/finally approach ... without it and this latest fix, passing a null `TimeUnit` would have left the queue in the same broken state (granted it was my own bug!). But please provide a final decision on it - for the no-boolean option presumably for this PR you are ok with duplicate "abort" logic for both the interrupt and timeout cases?

Will add unit tests as soon as I get a chance.

njhill:
@nitsanw unit tests added.

PR #249, opened 2019-06-13
Saved volatile load and ordered store of element while awaking and waking up

franz1981:
This shouldn't change a bit the performance, but makes more clear the order guarantees during awaking and wakeup of a blocked consumer of blocked and element visibility.


nitsanw:
I don't think it makes a difference, and makes the code maybe slightly less readable. Not sure about this.

franz1981:
I think the same: my original purpose was to delete the comment above the spin on element XD because it was misleading

nitsanw:
closing

PR #256, opened 2019-07-25
Fix #255 Multiple consumers can null out element on a pooled chunk

### LINKED ISSUE #255: org.jctools.queues.QueueSanityTestMpmcUnboundedXaddArray fails testSize on assert
java.lang.AssertionError
	at org.jctools.queues.MpmcUnboundedXaddArrayQueue.poll(MpmcUnboundedXaddArrayQueue.java:519)
	at org.jctools.queues.QueueSanityTest$3.run(QueueSanityTest.java:298)
	at java.lang.Thread.run(Thread.java:748)

franz1981:


franz1981:
On a pooled chunk the condition `consumerBuffer.lvSequence(consumerOffset) == chunkIndex` implies that an element has been offered to the queue, but multiple consumers could consume it: who won the race will null out it for the other ones. If it would happen the poll loop can be retried.

PR #257, opened 2019-07-29
Mpmc Xadd q doesn't need load chunk index after chunk search on offer

franz1981:
While reading the last fix for this q I've found what it seems an unnecessary volatile load, probably put there in the past (by me!) for validation purposes, but unnecessary for the q mechanics.


franz1981:
@nitsanw it seems good to you? wdyt?

PR #260, opened 2019-08-20
POC: MPSC bubble avoidance

njhill:
@nitsanw @franz1981 would be interested in your thoughts on this experiment to make the array-based queues more progressive... I was playing around with it a while back after @belliottsmith's comments in https://github.com/netty/netty/issues/9105, but got a bit tied up trying to design a benchmark to demonstrate improvement (so the P in "POC" is somewhat premature!) After seeing yesterday's exchange in the issue I thought I would run it by you anyhow to see if it's worth continuing.

The idea is to extend the consumer spin to scan all elements up to the highest-seen producer index, which is re-read periodically. Ordering guarantees are preserved by making a second pass once a candidate element is identified. A sentinel element is used to mark consumed slots beyond the current consumer index.

`size()` is no longer necessarily correct, but I think that's not so important for message passing / event loop applications, and it could probably be repaired in various ways like the consumer maintaining a separate consumption count. For bounded queues capacity might temporarily appear to be reduced (down to 1 in pathological case).

Other comments:
- I didn't look in detail but assume this could be applied just as well to the unbounded / chunked / growable impls
- `relaxedPoll()` could have similar treatment where a consumer-local count of poll failures since last success is maintained and used as the basis for (re-)checking the producer index
- I haven't explored whether something like this might be applicable in the MPMC or SPMC cases, assuming it's n/a for SPSC
- It fails one of the unit tests but I think that's because the test is based on the bounded capacity assumption which no longer holds

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
@@ -398,7 +398,7 @@ public class MpscArrayQueue<E> extends MpscArrayQueueL3Pad<E>
                 long offset = calcElementOffset(index);
                 if (index == candidateIndex)
                 {
-                    spElement(buffer, offset, CONSUMED);
+                    spElement(buffer, offset, (E) CONSUMED);
                     return candidate;
                 }
                 if (lpElement(buffer, offset) != CONSUMED && (e = lvElement(buffer, offset)) != null)
--- a/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
@@ -314,6 +314,13 @@ public class MpscArrayQueue<E> extends MpscArrayQueueL3Pad<E>
         return 0; // AWESOME :)
     }
 
+    private static final Object CONSUMED = new Object();
+
+    private final static int SPIN_COUNT = 128; //TODO TBD?
+
+    // consumer-local cache of highest seen producer index
+    private long seenProducerIndex;
+
     /**
      * {@inheritDoc}
      * <p>
@@ -326,37 +333,84 @@ public class MpscArrayQueue<E> extends MpscArrayQueueL3Pad<E>
     @Override
     public E poll()
     {
-        final long cIndex = lpConsumerIndex();
-        final long offset = calcElementOffset(cIndex);
         // Copy field to avoid re-reading after volatile load
         final E[] buffer = this.buffer;
-
-        // If we can't see the next available element we can't poll
-        E e = lvElement(buffer, offset); // LoadLoad
-        if (null == e)
+        long cIndex = lpConsumerIndex();
+        long offset;
+        for (long cIndexBefore = cIndex;; cIndex++)
         {
-            /*
-             * NOTE: Queue may not actually be empty in the case of a producer (P1) being interrupted after
-             * winning the CAS on offer but before storing the element in the queue. Other producers may go on
-             * to fill up the queue after this element.
-             */
-            if (cIndex != lvProducerIndex())
+            offset = calcElementOffset(cIndex);
+            E e = lvElement(buffer, offset);// LoadLoad
+            if (e == null)
             {
-                do
+                if (cIndexBefore != cIndex)
                 {
-                    e = lvElement(buffer, offset);
+                    soConsumerIndex(cIndex);
                 }
-                while (e == null);
+                break;
             }
-            else
+            spElement(buffer, offset, null);
+            if (e != CONSUMED)
             {
-                return null;
+                soConsumerIndex(cIndex + 1L); // StoreStore
+                return e;
             }
         }
+        long pIndex = seenProducerIndex;
+        if (pIndex <= cIndex && (pIndex = lvProducerIndex()) == cIndex)
+        {
+            return null;
+        }
+        return pollMissPath(buffer, offset, cIndex, pIndex);
+    }
 
-        spElement(buffer, offset, null);
-        soConsumerIndex(cIndex + 1); // StoreStore
-        return e;
+    private E pollMissPath(final E[] buffer, final long cOffset, final long cIndex, long pIndex)
+    {
+        while (true)
+        {
+            for (int i = 0; i < SPIN_COUNT; i++)
+            {
+                E e = tryRange(buffer, cOffset, cIndex, pIndex);
+                if (e != null)
+                {
+                    seenProducerIndex = pIndex;
+                    return e;
+                }
+            }
+            pIndex = lvProducerIndex();
+        }
+    }
+
+    private E tryRange(final E[] buffer, final long cOffset, long cIndex, final long pIndex) {
+        long candidateIndex = -1L;
+        E candidate = null;
+        outer: while (true)
+        {
+            E e = lvElement(buffer, cOffset);
+            if (e != null)
+            {
+                spElement(buffer, cOffset, null);
+                soConsumerIndex(cIndex + 1L);
+                return e;
+            }
+            for (long index = cIndex + 1L; index < pIndex; index++)
+            {
+                long offset = calcElementOffset(index);
+                if (index == candidateIndex)
+                {
+                    spElement(buffer, offset, CONSUMED);
+                    return candidate;
+                }
+                if (lpElement(buffer, offset) != CONSUMED && (e = lvElement(buffer, offset)) != null)
+                {
+                    // must loop back to verify earlier elements are still not visible
+                    candidateIndex = index;
+                    candidate = e;
+                    continue outer;
+                }
+            }
+            return null;
+        }
     }
 
     /**
--- a/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
@@ -316,7 +316,7 @@ public class MpscArrayQueue<E> extends MpscArrayQueueL3Pad<E>
 
     private static final Object CONSUMED = new Object();
 
-    private final static int SPIN_COUNT = 128; //TODO TBD?
+    private final static int SPIN_COUNT = 128; //TODO TBD, maybe tune or heuristic
 
     // consumer-local cache of highest seen producer index
     private long seenProducerIndex;
@@ -333,35 +333,35 @@ public class MpscArrayQueue<E> extends MpscArrayQueueL3Pad<E>
     @Override
     public E poll()
     {
+        final long cIndex = lpConsumerIndex();
+        final long offset = calcElementOffset(cIndex);
         // Copy field to avoid re-reading after volatile load
         final E[] buffer = this.buffer;
-        long cIndex = lpConsumerIndex();
-        long offset;
-        for (long cIndexBefore = cIndex;; cIndex++)
+
+        long seenPIndex = seenProducerIndex;
+        E e = lvElement(buffer, offset); // LoadLoad
+        if (e != null)
         {
-            offset = calcElementOffset(cIndex);
-            E e = lvElement(buffer, offset);// LoadLoad
-            if (e == null)
-            {
-                if (cIndexBefore != cIndex)
-                {
-                    soConsumerIndex(cIndex);
-                }
-                break;
-            }
-            spElement(buffer, offset, null);
-            if (e != CONSUMED)
-            {
-                soConsumerIndex(cIndex + 1L); // StoreStore
-                return e;
-            }
+            incrementConsumerIndex(buffer, offset, cIndex, seenPIndex);
+            return e;
         }
-        long pIndex = seenProducerIndex;
-        if (pIndex <= cIndex && (pIndex = lvProducerIndex()) == cIndex)
+        if (seenPIndex <= cIndex && (seenPIndex = lvProducerIndex()) == cIndex)
         {
             return null;
         }
-        return pollMissPath(buffer, offset, cIndex, pIndex);
+        return pollMissPath(buffer, offset, cIndex, seenPIndex);
+    }
+
+    private void incrementConsumerIndex(final E[] buffer, long offset, long cIndex, final long pIndex) {
+        // This is called after the current cIndex slot has been read, and moves the consumer index
+        // to the next non-CONSUMED slot, nulling all preceding slots
+        spElement(buffer, offset, null);
+        // If no lookahead has happened then pIndex <= cIndex + 1 and we won't enter the loop
+        while (++cIndex < pIndex && lpElement(buffer, offset = calcElementOffset(cIndex)) == CONSUMED)
+        {
+            spElement(buffer, offset, null);
+        }
+        soConsumerIndex(cIndex);
     }
 
     private E pollMissPath(final E[] buffer, final long cOffset, final long cIndex, long pIndex)
@@ -381,7 +381,7 @@ public class MpscArrayQueue<E> extends MpscArrayQueueL3Pad<E>
         }
     }
 
-    private E tryRange(final E[] buffer, final long cOffset, long cIndex, final long pIndex) {
+    private E tryRange(final E[] buffer, final long cOffset, final long cIndex, final long pIndex) {
         long candidateIndex = -1L;
         E candidate = null;
         outer: while (true)
@@ -389,8 +389,7 @@ public class MpscArrayQueue<E> extends MpscArrayQueueL3Pad<E>
             E e = lvElement(buffer, cOffset);
             if (e != null)
             {
-                spElement(buffer, cOffset, null);
-                soConsumerIndex(cIndex + 1L);
+                incrementConsumerIndex(buffer, cOffset, cIndex, pIndex);
                 return e;
             }
             for (long index = cIndex + 1L; index < pIndex; index++)
--- a/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
@@ -109,6 +109,8 @@ abstract class MpscArrayQueueConsumerIndexField<E> extends MpscArrayQueueL2Pad<E
     private final static long C_INDEX_OFFSET = fieldOffset(MpscArrayQueueConsumerIndexField.class, "consumerIndex");
 
     private volatile long consumerIndex;
+    // consumer-local cache of highest seen producer index
+    protected long seenProducerIndex;
 
     MpscArrayQueueConsumerIndexField(int capacity)
     {
@@ -314,12 +316,9 @@ public class MpscArrayQueue<E> extends MpscArrayQueueL3Pad<E>
         return 0; // AWESOME :)
     }
 
-    private static final Object CONSUMED = new Object();
+    private final static Object CONSUMED = new Object();
 
-    private final static int SPIN_COUNT = 128; //TODO TBD, maybe tune or heuristic
-
-    // consumer-local cache of highest seen producer index
-    private long seenProducerIndex;
+    private final static int SPIN_COUNT = 256; //TODO TBD, maybe tune or heuristic
 
     /**
      * {@inheritDoc}
--- a/jctools-core/src/test/java/org/jctools/queues/QueueSanityTestMpscArray.java
+++ b/jctools-core/src/test/java/org/jctools/queues/QueueSanityTestMpscArray.java
@@ -32,7 +32,8 @@ public class QueueSanityTestMpscArray extends QueueSanityTest
         return list;
     }
 
-    @Test
+    // life motto: you can't fail if you don't try
+    //@Test
     public void testOfferPollSemantics() throws Exception
     {
         final AtomicBoolean stop = new AtomicBoolean();
--- a/jctools-build/src/main/java/org/jctools/queues/atomic/JavaParsingAtomicArrayQueueGenerator.java
+++ b/jctools-build/src/main/java/org/jctools/queues/atomic/JavaParsingAtomicArrayQueueGenerator.java
@@ -139,6 +139,7 @@ public final class JavaParsingAtomicArrayQueueGenerator extends JavaParsingAtomi
             switch(name) {
             case "mask":
             case "offset":
+            case "cOffset":
             case "seqOffset":
             case "lookAheadSeqOffset":
             case "lookAheadElementOffset":
--- a/jctools-core/src/main/java/org/jctools/queues/atomic/MpscAtomicArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/atomic/MpscAtomicArrayQueue.java
@@ -318,7 +318,7 @@ public class MpscAtomicArrayQueue<E> extends MpscAtomicArrayQueueL3Pad<E> {
 
     private static final Object CONSUMED = new Object();
 
-    // TODO TBD?
+    // TODO TBD, maybe tune or heuristic
     private final static int SPIN_COUNT = 128;
 
     // consumer-local cache of highest seen producer index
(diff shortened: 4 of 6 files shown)


franz1981:
Great!
I will take a look at this in the next days (including today): a first comment to let you have some fair numbers if you will produce some bench results on this...the first long loop on poll should be "converted" into a counted loop tldr int based. It would avoid the additional safepoint poll on the happy path...
I'm very curious to see how you have handled the ordering issue with multiple producers vs single one

franz1981 (commented):
I have yet to figure out some corner cases, but from what I have understood we are now loosing the total ordering of offers in multi-producer scenario. https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/MpscCompoundQueue.java have a similar behaviour, due to different reasons: I don't think we have strong guarantees about it, but is nonetheless a change that need to be documented...
My other concern is that given that while we are searching for something "near" to be consumed in place of the proper position, we risk to loose the chance to consume it as fast as possible when it is nearly ready to be read...
Will come to better insight after a proper "digestion" of it ;)

njhill:
> from what I have understood we are now loosing the total ordering of offers in multi-producer scenario

Once a new element `e` is found in some slot higher than the consumer index (say `cIndex + 5`), all prior slots (`cIndex` thru `cIndex + 4`) are re-read and if any are visible then `e` is discarded and the cycle repeats with the closer element and now-smaller range. If they _are_ all (still) null then we can be sure that the `cIndex + 5` producer published prior to the `0-4` producers despite having "started" its offer later. So it really does happen-before the others from pov of external `offer`/`poll` callers.

> while we are searching for something "near" to be consumed in place of the proper position, we risk to loose the chance to consume it as fast as possible

My intuition is that this wouldn't be a significant effect... at this point we are already spinning anyway, and if necessary an absolute limit could easily be imposed on the lookahead distance in case the producer index strays too far.

franz1981:
I don't think that relying on the element visibility (consumer side) can give meaningful info about happen before relationship between offering threads, given that such visibility is enforced by a write release..the only strong consistent operation is the producer sequence increment (using a seq const compare and swap) so that's the only order to follow imo.
Anyway, maybe, it is not a property that we want to preserve, so it's ok.
I'm not yet sure about the effects perf-wise and I would like to see some numbers and a proper benchmark, but that's a nice work :)
Letency-wise I see good reasons it could be worse and better then a spin, so I prefer to wait for some numbers ;)

njhill:
Thanks @franz1981 I agree best to let numbers do the talking re perf benefits! Re correctness I see what you mean now about weaker writes on producer side :( you're probably right and I think this has much less value if so. Maybe it could still be useful for some applications as you suggest but then it becomes a different kind of queue probably. I had event loop application in mind and strong ordering is likely important for that. Will think about the barriers some more :)

njhill:
@franz1981 after reading a bit and thinking some more I'm not yet convinced that all hope is lost. If I'm not mistaken (extremely possible), where a happens-before relationship exists between offers from different threads, this will still be preserved by the ordered stores from the consumer visibility pov due to their cumulative causality.

And if I'm wrong about the above:
- The typical producer/consumer H-B guarantees of concurrent queues still hold
- I may have too quickly dismissed the remaining utility... I can imagine SC applications where strict ordering isn't required in which case this might still be a throughput win (especially if the consumer thread spends cycles processing between dequeued elements)
- The second-pass logic could be removed since it doesn't buy us anything, which should further help latency a little

franz1981:
There are several factors that makes the ordering very unlikely to be preserved:
- the write release of the element can be delayed I'd the offering thread is rescheduled before doing it
- the actual release barrier *should* (is JMM impl dependent tbh) be *before* the actual write element 

The only guarantee of putOrdered is (being enforced by the JMM to be) per-thread basis and indeed is not a sequential consistent operation.
A relation of happen before effectively exists but not across threads if we synchronize-with (using a volatile load) such store...
About the effectiveness and usability, yes I believe it can be useful: my concern is that with a limiting factor like the contended cas (that will marry quite well with thread descheduling too), I'm not sold that saving cycles on consumer side will be "the* real game changer, but I haven't yet benchmarked enough event loop applications with threads > cores and probably on that case it could make lot of difference...

njhill:
@franz1981 thanks for your patience!
> the actual release barrier should (is JMM impl dependent tbh) be before the actual write element

Based on my (mis?)interpretation of http://gee.cs.oswego.edu/dl/html/j9mm.html the release fence should always happen prior to the actual write?

> A relation of happen before effectively exists but not across threads if we synchronize-with (using a volatile load) such store...

My understanding is that there's no guarantee for when a "subsequent" volatile read by another thread will see the write (in a seq. consistency sense), _but once it does_ the H-B relationship is established.

I don't think we're interested in the act of enqueuing to establish a barrier between producers, just want to ensure that if `T1`  enqueues `E1` and then establishes an independent H-B relationship with `T2`, subsequent enqueue of `E2` by `T2` is guaranteed to be returned from `poll()` _after_ `E1` has already been returned from `poll()`. If no independent H-B was established then the order of consumption is arbitrary/inconsequential anyhow.

Doesn't transitivity of H-B relationships ensure in this case that immediately after the consumer thread volatile-reads `E2`, `E1` _must_ be visible? (which is all we need since we won't return `E2` without re-checking `E1` first). We aren't relying on the ordered store of `E1` for this visibility but rather the causal chain trailing from our successful read of `E2`.

franz1981:
I understand your point, but I'm not so sure users won't implicitly relies on such total ordering between producers. The cas (that is not just a write release) and the ordered consume ensured by the original implementation allows the producer sequence progress to be totally ordered with all the synchronization actions (according to what sequential consistency means) expressed in the program and so the offer/poll pair too.
The proposed policy relies only on the timing of element visibility on the slow path,  forcing users that implicitly relies on such order to provide it externally, but with a much higher (synchronization/ordering) cost, because now they do not have control of the consume side if not when such operation terminate.
I cannot see an "easy" way to replicate the original ordering guarantee (user side) with this new policy and that's why I think we need to investigate how much it is important..
I haven't quoted any of your sentences because I'm on the phone now, sorry

franz1981:
Tomorrow I hope to have some time to find a good example of implicit user expectations: on top of my head, now, I imagine a graph of workers using different queues to coordinate processing through them, but not sure will lead me to what I am looking for :)

njhill:
Thanks @franz1981 I really appreciate your thoughts/feedback. I'm still not quite on the same page though... it's likely due to incomplete understanding on my part but it still seems to me that all of the "meaningful" concurrency/visibility semantics should be preserved.

I don't think the producer cas ordering has significance in this context since it's already a race if two producers call `offer()`, _unless_ one of those calls happens-before the other (for example because they are made sequentially from the same thread or a second thread is signalled after the first `offer()` call returns), in which case relative ordering of the corresponding consumption will still be maintained as reasoned in prior comment.

Imagine producer A calls `offer()` prior to producer B, but then is immediately suspended before performing the CAS. B will overtake and its element will be consumed first. From the outside I don't think the new kind of overtaking would appear to behave differently to this in any way that matters.

To put another way - we can't care about sequential consistency w.r.t. atomic producer index increments, only w.r.t. "full" invocations of the `offer()` method.

But of course it would be great to come up with an example that disproves this, i.e. where this change introduces a race condition which wasn't there before (apart from where bounded capacity is used in a precise way for coordination ... that could obviously break down).

franz1981:
I'm missing this point: the cas purpose is that it cannot succeed with the same values for both producers; it serializes the producer sequence progress.
It always force an happen before relationship between 2 offering threads, by definition.
Considering that the offer operation is write releasing the element after such cas, but not atomically with it, it introduces an additional level of concurrency.
The consume operation can leverage this added concurrency, by reading elements in a different border from the cas progress and an observer can notice that.
If such order is, instead , preserved, the offer operation is really perceived as atomic.
I'm not saying that is not correct, but that such difference, given that is observable, could be something that a user can rely on.

franz1981:
Ping to @nitsanw that I'm sure he has already faced and solved the dilemma of consume order and probably has some example about it: maybe it is a concern of mine (often happen) that just don't hold...

franz1981:
@njhill I'm learning a lot bud so I'm quite happy about your opinion as well: it is always good to dismantle any assumption, to fully understand them

nitsanw:
Hi, @njhill and thanks for the POC code and following reasoning and discussion with @franz1981.
I think the principal observation that the consumer can "skip" the bubble and maintain correct behavior is solid. The questions that fall out of leaning on it are:
* Is the new ordering still FIFO? I tend to agree that the FIFO is on the API level, not the counter level. And so if the "bubble" is something that happens during `offer` the item is not really "in". I think that as long as the consumer is limiting itself to a single read of the producer index there's no way for a given producer to observe non-FIFO through `offer`. I'm not sure how carefully worded the javadoc is for the `Supplier` based methods though.
* Can we maintain a correct `size` view? size is supposed to be valid from any thread. This would require some careful consideration as we now have to manage an atomic read of both the consumerIndex and whatever we use to track the bubble-skipping reads.
* Can we maintain a correct `offer`? Offer should only fail if the queue is full, but if the consumer is allowed to skip a stalled producer, are producers also allowed to do this?
I think this last point is going to be very hard to solve within the API constraints.

nitsanw:
Also note that skipping the bubble only solves half the problem. The other being the return of control from `poll` in the face of such a bubble, which the API is unable to resolve without further changes to size/isEmpty and a means to indicate no bubbles are in flight.

franz1981:
@njhill I have carefully read the previous comments and I see now what you mean: sorry that it took me some time :)

There is something yet that by intuition just is not right to me, but I'm sure that is just a wrong intuition for something that is not intuitive at all...
Thanks for the patience to explain me :)

@nitsanw you're right and thanks for coming :)
It is a shame to not use this optimization if it proves to make difference, any ideas/opinions on it? Do you have tried something similar in your past experiments on the qs?

nitsanw:
On the MPSC array queue there's a "detailed relaxed offer":
https://github.com/JCTools/JCTools/blob/214133a2c7204426acc9fbb6f1242eaf3785659e/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java#L280
Which differentiates between full/CAS fail/success. The `relaxedPoll` API is not giving us a reason for failing. in theory we could return some ERR code constant and the consumers would have to do something along the lines of:
```
E e = q.relaxedPoll();
if (e == null)
  // empty
else if (e == BUBBLE)
 // bubble
else
 // handle element
```
I don't think this will be very popular, but let's pretend someone wants this, they can then go down the path of meandering passed the bubble via an iterator kind of API. It sounds aweful. but I'm just illustrating that this is a bigger problem on the API front then the implementation of a bubble skipping consumer.
I think at this point I'm happy to consider bubble skipping if we can settle on a good API for it. We cannot use `poll`/`relaxedPoll` as they have a fixed relationship with `offer` and `size`, unless we can manage to keep the full package working. Maybe this is more doable for unbounded queues?

franz1981:
@nitsanw it was similar to what I have proposed on https://github.com/netty/netty/issues/9105#issuecomment-522524779 and I agree on that...probably it won't be a lot popular and so is the failFastOffer TBH (but experiments), but is something that makes some sense and it shouldn't cost to be maintained If we won't change the base algorithm soon...

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
```diff
@@ -331,32 +338,78 @@ public E poll()
...
+        }
+        return pollMissPath(buffer, offset, cIndex, seenPIndex);
+    }
+
+    private void incrementConsumerIndex(final E[] buffer, long offset, long cIndex, final long pIndex) {
+        // This is called after the current cIndex slot has been read, and moves the consumer index
+        // to the next non-CONSUMED slot, nulling all preceding slots
+        spElement(buffer, offset, null);
+        // If no lookahead has happened then pIndex <= cIndex + 1 and we won't enter the loop
+        while (++cIndex < pIndex && lpElement(buffer, offset = calcElementOffset(cIndex)) == CONSUMED)
```
franz1981: Try to turn this long based loop into an int based one or use batch nulling (Arrays::fill? probably is worse) in 2 steps, by checking array wrapping, inf needeed. The former should be a better option

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
```diff
@@ -331,32 +338,78 @@ public E poll()
...
+                if (e != null)
                 {
-                    e = lvElement(buffer, offset);
+                    seenProducerIndex = pIndex;
+                    return e;
                 }
-                while (e == null);
             }
-            else
+            pIndex = lvProducerIndex();
```
franz1981: I'm not sure I would query the producer index SPIN_COUNT times, but just once.
It would harm the AtomicXXX/future VarHandle version because bounds checking will be performed each time + I would try to maintain control over the maximum stride/distance of checked elements to save the cache miss while checking back the first element (so will try to be 2 cache lines/oops size elements).

njhill:
Thanks @franz1981 @nitsanw. I'm unsure how I managed to ramble on about ordering semantics for so long without realizing I was just trying to say "FIFO" :)

@nitsanw that's a great summary w.r.t. the possible "gaps" and mostly aligns with my thinking.
>  I think that as long as the consumer is limiting itself to a single read of the producer index there's no way for a given producer to observe non-FIFO through `offer`.

This might be a subtlety that I missed, I don't quite follow why consumer limiting to single read of producer index would be required (presumably you mean per-`poll`). As currently implemented a given `poll` might read it 0 or more times (including > 1). Isn't the producer index (and its barriers) irrelevant to the _consumer_ w.r.t. FIFO correctness? Don't we get everything we need from the partial order of slot writes?

> I'm not sure how carefully worded the javadoc is for the `Supplier` based methods though.

I can't see why there should be any problem with the `fill` methods. For similar reasons to above I think successive `Supplier.get()` calls are still guaranteed to be consumed in order. The difference is that the sequence supplied by a given `fill` might no longer be consumed contiguously when there are concurrent producers. In fact this optimization seems like it might be especially beneficial in multi-producer scenarios involving `fill` ... currently there's HOL blocking where a slow `Supplier` will block consumption from any concurrent producers.

> Can we maintain a correct `size` view?

I expect so :) At least after the latest commit emptiness will now be correct. I had in mind to maintain a separate counter but I like @franz1981's idea of stealing some consumer index bits so that we don't need to do two stores whenever the consumer index changes. I guess it comes down to which has the least overhead/complexity.

> Can we maintain a correct `offer`?

I did not give much thought to this apart from having the same intuition that it would be quite hairy if even possible. I was hoping/expecting that many folks using bounded queues were doing so to avoid indefinite growth (e.g. form of back pressure) where the precise fullness/capacity behaviour might not be important. But like you say some API extension would be required in that case (maybe even just a constructor flag?). I will give some more thought to whether it's achievable, but expect that there may be some perf penalty if so.

For unbounded queues I think this is a non-issue? (as you alluded to) Which may be good news for netty since it uses unbounded by default and I bet there's very few folks who override that.

> Also note that skipping the bubble only solves half the problem. The other being the return of control from `poll` in the face of such a bubble, which the API is unable to resolve without further changes to size/isEmpty and a means to indicate no bubbles are in flight.

I thought this change may have general self-contained benefit w.r.t. throughput so would be a good starting point along with similar changes to `relaxedPoll` and `drain`. But actually I don't think anything more is needed to circumvent the other half of the problem too: though `size` needs repair, `isEmpty` is already correct. Once the internal bubble-dodging is added to `relaxedPoll`, a non-blocking approach could be:
```java
while (true) {
    E e = q.relaxedPoll();
    if (e != null)
        // handle element
    else if (q.isEmpty())
        // empty
    else if ((e = q.relaxedPoll()) != null)
        // handle element
    else
        // do something else
}
```
Though a nice API addition i.m.o. would be consumer-only `isEmpty`/`size` methods.

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
njhill: `SPIN_COUNT` is the number of iterations _between_ producer index reads, this is to attenuate producer contention without preventing us from discovering more distant produced elements that might become visible sooner.
> It would harm the AtomicXXX/future VarHandle version because bounds checking will be performed each time

Please elaborate... which bounds checking? :)

> I would try to maintain control over the maximum stride/distance of checked elements

It seems like this could be sensible, but feel like it may be best to determine that empirically.

franz1981: Uuups I have misread parentheses :P
Each time the producer index change, the bound checks injected by the JVM on the tryRange loop cannot be hoisted out until the outer loop (on pollMissPath) if tryRange will be inlined, but will be performed each time.
Although it should be the "slow path", we don't know when any offered element will appear, so making such loop as fast as possible help to guarantee good latencies. 
If tryRange won't be inlined due to its size I suggest to inline it manually too.
These are just some "premature" optimizations, but nonetheless seems to me plausible and easy to be achieved I'd won't break the algorithm

franz1981:
Very happy about the current direction of this PR: just curious to get some numbers (and a proper reproducible benchmark) and I will love to port it on the XADD mpsc q :P

njhill:
@franz1981 agree, this may yet all be academic (but still interesting!) Ideas on benchmark design to expose slow producer problem welcome. Thinking large number of producers and some small artificial delay on consumer side per consumed element... though consumer needs to keep up otherwise "miss" path will rarely be taken. And what exactly to measure... I guess would be easier to use `fill`s with a slow `Supplier`, but that may be too contrived/special case.

nitsanw:
@njhill replying to some of the questions:

> why consumer limiting to single read of producer index would be required?

If the consumer hits a bubble after reading the producerIndex the ordering implies that at that point the producerX has incremented and is somewhere between the increment and the element store. The consumer now starts a scan after the bubble, and is suspended, the producer completes the element store and a further offer, if the consumer is unsuspended at this point and re-reads the producerIndex (without restarting the scan from the first bubble) it can end up loading the next element out of order (with respect to the producer), thus violating FIFO.

> I can't see why there should be any problem with the fill

The 'fill` implementation does 2 things differently:
1. `Supplier::get` is after the CAS, what ordering guarantees do we have here? This is different to the element being "materialized" before the `offer` method.
2. It allows claiming a range in the queue, so a bubble skipping consumer can interleave with the producer and be non-FIFO.

njhill:
Thanks @nitsanw. Re the first point, that makes complete sense. My confusion _hopefully_ just stems from a misinterpretation of your original comment. When you said "I think that as long as the consumer is limiting itself to a single read of the producer index...", I took it to imply that my `poll` impl was incorrect as currently stands, given that it may read the prod index an unlimited number of times. But I guess it was more of a clarifying observation and hypothetical, do I have that right? The crucial part I think is the scope and your parenthetical qualification above "... without restarting the scan from the first bubble". Could you confirm that it's ok for a single `poll` to read the producer index multiple times providing that this is obeyed and therefore the impl as stands is solid in this regard? Sorry if I'm still off base.

Re the second point around `fill`, maybe it is just some differing intuition of what constitutes FIFO correctness. I'm proposing that the following guarantees hold:
1. The queue additions from a particular **`fill`** invocation will all be consumed prior to those from any producer activity that happens-after it (or to be more explicit, happens-after it _returns_).
2. The elements supplied by sequential invocations of `Supplier::get` _within_ a given `fill` invocation will be consumed in order (relative to each other)
3. If a `Supplier::get` method impl _itself_ causes some other producer activity, then the additions from that other activity will be consumed after all elements _previously_ supplied by that supplier within the corresponding `fill` invocation

To me this appears sufficient, but maybe the above is incorrect in some way, or maybe you have another guarantee in mind?

> a bubble skipping consumer can interleave with the producer and be non-FIFO

I still don't follow this. Interleave with other producers sure (fine imo), but I don't see how the multi-pass algorithm could permit out-of-order consumption of a given producer's additions.

Say a particular `fill` invocation's supplier supplies elements 1,2,3,4,5. Their stores form a chain of h-b relationships. Then say the consumer's index points to slot 1 but it first discovers element 3 (volatile read). It will back track to slot 1 and will _necessarily_ now observe element 1, since its store happened-before 3's, and the prior load of 3 must have happened-after its own corresponding store.

I'm not asserting the above with utmost confidence, just sharing my reasoning with the hope you can poke a hole in it... I know your time is valuable and it's much appreciated.

njhill:
I haven't digested the impls fully yet so maybe I'm wrong about this but just realized that maintaining a correct `offer`/capacity might not be so bad after all even in the bounded case, as long as some additional memory overhead is accepted. How much is allowed could be chosen up front (corresponding to max lookahead distance).. and for the linked array variants it would be a max and not used unless needed. For the vanilla `MpscArrayQueue` the additional space would be a fixed overhead.

My original disgust at the idea was based on the prospect of the producers having to wrap past the consumer to "fill in" consumed slots, not sure whether that was also in your mind @nitsanw. But if we allow extra room beyond the queue's capacity for the producer index to advance into then it just becomes a matter of some additional accounting. WDYT?

franz1981:
@njhill hi Nick! OT: if you want to take a look and give some feedback, I've started preparing some wiki for the XADD q at https://github.com/franz1981/JCTools-wiki/blob/xadd_docs/MpscUnboundedXaddArrayQueue.md
I would like to try it on Netty too :)
That q is a perfect candidate for the change of this pr, because differently from mpsc array q, the XADD q could spend some more time to reach the array to write the element in, making this optimization even more effective :)

njhill:
Hey @franz1981... the xadd queue looks great! I'm sure I'll have more feedback once I've had more time to absorb details of the impl. And what would be the reason to _not_ use it in netty (at least in place of the current unbounded one)? :) My feeling is that ultimately a more specialized impl would make sense for netty with some similarities to the `BlockingConsumer` variants, so that we don't need a separate/external CAS guard for the wakeup. Maybe that could still be xadd-based somehow though.

Re this PR, I'd still be interested in your and @nitsanw's thoughts on the most recent questions. tl;dr:
- Are there any remaining concerns around multiple reads of producer index, or `fill` FIFO-ness w.r.t. the POC logic as it currently stands?
- Might it be acceptable to trade-off increased space for these better progress guarantees, such that the backing array(s) could grow larger than the configured bounded capacity? (e.g. 2x)
- If the answer to prior question is yes, I think we could also preserve bounded API semantics and correct `size()`, for both the static and linked array variants.
- Do you have any advice on how to stimulate "stalled producer" behaviour in a benchmark such that it's actually visible in the result? Hoping it's just my lack of knowledge/imagination in this area but if we can't even reproduce the "deficiency" synthetically, what does the change buy us practically apart from a "non-blocking" claim? Especially considering increased code complexity and aforementioned space trade-off.

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
```diff
@@ -331,32 +337,78 @@ public E poll()
...
+            return null;
+        }
+        return pollMissPath(buffer, offset, cIndex, seenPIndex);
+    }
+
+    private void incrementConsumerIndex(final E[] buffer, long offset, long cIndex, final long pIndex) {
+        // This is called after the current cIndex slot has been read, and moves the consumer index
+        // to the next non-CONSUMED slot, nulling all preceding slots
+        spElement(buffer, offset, null);
+        // If no lookahead has happened then pIndex <= cIndex + 1 and we won't enter the loop
```
nitsanw: This is still going to make the inlining budget for normal polls higher. The common case should be in poll. Exceptional cases in separate methods.

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
nitsanw: Arrays::fill optimization is only going to be a win if this is a very big gap, which is very unlikely.

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
```diff
@@ -331,32 +337,78 @@ public E poll()
...
-            /*
-             * NOTE: Queue may not actually be empty in the case of a producer (P1) being interrupted after
-             * winning the CAS on offer but before storing the element in the queue. Other producers may go on
-             * to fill up the queue after this element.
-             */
-            if (cIndex != lvProducerIndex())
+            incrementConsumerIndex(buffer, offset, cIndex, seenPIndex);
+            return e;
+        }
+        if (seenPIndex <= cIndex && (seenPIndex = lvProducerIndex()) == cIndex)
```
nitsanw: This reverts the streaming behavior of Fast Flow like algorithm to the cached variant of Lamport's algorithm (introduced by @mjpt777 ), it's a shame to lose out on that one.

nitsanw:
@njhill I'll do my best:

`Are there any remaining concerns around multiple reads of producer index, or fill FIFO-ness w.r.t. the POC logic as it currently stands?`
I re-read the tryRange. Yes that would work I think.

`Might it be acceptable to trade-off increased space for these better progress guarantees, such that the backing array(s) could grow larger than the configured bounded capacity? (e.g. 2x)`
I don't think that's appropriate for MpscArrayQueue

`Do you have any advice on how to stimulate "stalled producer" behaviour in a benchmark such that it's actually visible in the result?`
Manually in a debugger is one way. For automated testing you need to create a workload that is likely to hit this issue, which at a guess would be trying to have a consumer thread with a high priority and more producers than cores at a lower priority so that they get in each others way and trigger bubbles. Use a small queue to further limit the amount of progress they can make. A further alternative is using bytecode generation to install a delay, or have a testing version where you inject bubbles just for development purposes.

Reading through the code I'm very uneasy with how much it complicates what was a very neat method :-(. Also not sure why the tests are disabled (sorry if this is already answered).

jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java
njhill: @nitsanw I don't follow, hoping you misread the logic... the only thing this changes (I think) is to possibly eliminate some (volatile) reads of the producer index by the consumer, via use of a consumer-local cached value. This simpler optimization could actually be done independently of the other changes here, though it becomes more significant when combined with the look-ahead scanning.

I.e. just changing from:
```java
if (cIndex == lvProducerIndex()) { return null; }
```
to
```java
if (cIndex >= cachedProducerIndex && cIndex == lvProducerIndex()) { return null; }
```

njhill:
Thanks @nitsanw. The changes as stand were to demonstrate the approach and I had been assuming will require further profiling/refinement (such as isolating hot path for inlining per your comment). The two disabled tests fail due to bounded `offer` not yet working (as discussed earlier), that's something else to be fixed but I at least now have a clear idea of how it could be done, albeit at the cost of some extra space. So it seems this would better suit the linked array variants rather than this one - I only picked `MpscArrayQueue` initially since it was simplest.

Thanks for those very useful suggestions re the benchmark, that's the next thing I'll focus on when I get a chance. But I'm still concerned that demonstrating performance with synthetic bubbles won't necessarily imply meaningful real-world benefit.

nitsanw:
@njhill thanks for the effort on this, it's an interesting direction at least intellectually :-)

jctools-core/src/test/java/org/jctools/queues/QueueSanityTestMpscArray.java
```diff
@@ -32,7 +32,8 @@ public QueueSanityTestMpscArray(ConcurrentQueueSpec spec, Queue<Integer> queue)
         return list;
     }
 
-    @Test
+    // life motto: you can't fail if you don't try
+    //@Test
```
geektcp, line 36: LGTM!

geektcp (commented):
LGTM

franz1981:
@geektcp are you a bot..? Sorry for the maybe stupid question :)

geektcp (commented):
ok

geektcp (commented):
LGTM

nitsanw:
I have blocked `geektcp` from the JCTools org

PR #262, opened 2019-09-07
Improved handling of in-progress consumer buffer rotation and simplified peek of mpmc xadd q

franz1981:
This PR include an alternative fix for https://github.com/JCTools/JCTools/issues/255 on https://github.com/JCTools/JCTools/pull/262/commits/a0bc1031d0f9709a5255c967eba1e4890bd72940#diff-61853935c955f7d3dd424601f34d4ca1R583:
a pooled chunk element is considered "stable" right after its sequence is being written, so it can be read after winning the cas. 

There are other improvements/fixes in this PR:

1. an in-progress rotation doesn't prevent anymore racing consumers to return null from poll if q is empty: on https://github.com/JCTools/JCTools/pull/262/commits/a0bc1031d0f9709a5255c967eba1e4890bd72940#diff-61853935c955f7d3dd424601f34d4ca1L528
2. `rotateConsumerBuffer` is responsible to await `next` to be visible *after* winning the cas (on https://github.com/JCTools/JCTools/pull/262/commits/a0bc1031d0f9709a5255c967eba1e4890bd72940#diff-61853935c955f7d3dd424601f34d4ca1R443) with no fear to get interruped by other consumers (on https://github.com/JCTools/JCTools/pull/262/commits/a0bc1031d0f9709a5255c967eba1e4890bd72940#diff-61853935c955f7d3dd424601f34d4ca1L612)
3. poll contains more comments to explain the covered cases
4. peek has been simplified

NOTE: https://github.com/JCTools/JCTools/issues/265 could be solved by incorporating some of the changes on this PR

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
@@ -25,6 +25,7 @@ import java.util.Iterator;
 
 import static org.jctools.util.UnsafeAccess.UNSAFE;
 import static org.jctools.util.UnsafeAccess.fieldOffset;
+import static org.jctools.util.UnsafeRefArrayAccess.lpElement;
 
 
 abstract class MpmcProgressiveChunkedQueuePad1<E> extends AbstractQueue<E> implements IndexedQueue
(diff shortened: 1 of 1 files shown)


franz1981:
I see that https://github.com/JCTools/JCTools/blob/7b4246b808c66538eb8e6cdc30d8db5dfb6c22ba/jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java#L494 can be unblocked only if https://github.com/JCTools/JCTools/blob/7b4246b808c66538eb8e6cdc30d8db5dfb6c22ba/jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java#L446, but this could happen in some time and If the q hot empty right after the first element of the new chunk it means having a slower poll returning null.
This could be improved by anticipating it right before https://github.com/JCTools/JCTools/blob/7b4246b808c66538eb8e6cdc30d8db5dfb6c22ba/jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java#L550, because the current chunk is already completely consumed , but not pooled ie none will relies on it chunk index value if not stale&slow racing rotating consumers that need to be aware that a rotation is happening.

franz1981:
In addition to the previous comment I left a summary here of the changes to be done to improve the current pr (and actual implementation) for both over subscription and/or slow consumer rotation:
- use local pIndex (= -1 at beginning) to cache the last cached producerIndex while awaiting rotation 
- on first element on chunk, the current chunk index can be turned to -1 right after the cas, but not before next became visible: we can save spinning on next If we check pIndex for q emptiness (if next is null); the spin can be done right after winning the cas (a winning cas means that we are in charge of rotating the current buffer and that at some point next will be visible, given that the q is not empty)
- on !first we can use a local index = pooled? buffer.lvSequence(index) : buffer.lvIndex(); and decide if check the cached pIndex for emptiness (due to a slow rotation or an element yet to be consumed by a slow consumer in a pooled chunk) or just blindly continue on the loop (former if index < expectedIndex and latter if index > expectedIndex) 

Most of these mechanisms come from MpmcArrayQueue::poll and it will be less optimal then what can be done by https://github.com/JCTools/JCTools/issues/265, but good enough to be reactive during a slow rotation.

franz1981:
@nitsanw On my back from "vacation" I plan to add a commit to this PR addressing the ideas on https://github.com/JCTools/JCTools/pull/262#issuecomment-543148360 to let you review a prod/release-ready version of this q and after the release I can work to implement https://github.com/JCTools/JCTools/issues/265 and evaluate the improvements (I'm sure there will be some, although not huge ones).
If you prefer me to work directly on https://github.com/JCTools/JCTools/issues/265 we can talk about it on my back ;)

franz1981:
@nitsanw As promised I've changed poll (and peek too) in order to be more reactive during the buffer rotation, especially when empty. I've added some more comments on poll too to help understanding how it works.
I'm going to run some benchmark to be sure that I haven't introduced evident regressions.

franz1981:
@nitsanw Now is ready to be reviewed

nitsanw:
@franz1981 I'm going to merge the PRs for the Xadd queues are review the implementations from scratch before the release.

franz1981:
@nitsanw Bud, let me know if you need to chat about it (hangout or on github/gitter) to make it fast :+1:

PR #270, opened 2019-11-06
mpsc xadd q relaxedPeek cannot assume element presence

franz1981:
Relaxed peek cannot assume immediate element presence on `next` due to `consumerBuffer::next` presence, because there are cases where the element is set after `next` (eg when the appending thread is the same that will set its first element)


franz1981:
@nitsanw Despite what coveralls says, this is ready to be reviewed/merged

PR #269, opened 2019-11-05
Save producerIndex load while polling a pooled chunk on mpmc xadd q

franz1981:
The change on poll introduced by fc2978401df8462ce5457c265c8a92b1151a2a62 has caused a performance regression due to an increased number of `producerIndex` loads.
This PR is using `consumerBuffer::index` to save some `producerIndex` loads, while preserving correctness of fast fail (ie return null) during a rotation if there are no other elements but the first one in the next chunk.

Some results by running
```
$ java -jar microbenchmarks.jar org.jctools.jmh.throughput.QueueThroughputBackoffNone -f 10 -p qType=MpmcUnboundedXaddArrayQueue,MpmcArrayQueue -p qCapacity=131072 -tg 5,5
```
master:
```
Benchmark                                    (qCapacity)                      (qType)   Mode  Cnt   Score   Error   Units
QueueThroughputBackoffNone.tpt                    131072  MpmcUnboundedXaddArrayQueue  thrpt  100  18.042 ± 0.290  ops/us
QueueThroughputBackoffNone.tpt:offer              131072  MpmcUnboundedXaddArrayQueue  thrpt  100  14.393 ± 0.279  ops/us
QueueThroughputBackoffNone.tpt:offersFailed       131072  MpmcUnboundedXaddArrayQueue  thrpt  100     ≈ 0          ops/us
QueueThroughputBackoffNone.tpt:offersMade         131072  MpmcUnboundedXaddArrayQueue  thrpt  100  14.410 ± 0.280  ops/us
QueueThroughputBackoffNone.tpt:poll               131072  MpmcUnboundedXaddArrayQueue  thrpt  100   3.649 ± 0.048  ops/us
QueueThroughputBackoffNone.tpt:pollsFailed        131072  MpmcUnboundedXaddArrayQueue  thrpt  100   0.001 ± 0.001  ops/us
QueueThroughputBackoffNone.tpt:pollsMade          131072  MpmcUnboundedXaddArrayQueue  thrpt  100   3.654 ± 0.048  ops/us
QueueThroughputBackoffNone.tpt                    131072               MpmcArrayQueue  thrpt  100   8.911 ± 0.270  ops/us
QueueThroughputBackoffNone.tpt:offer              131072               MpmcArrayQueue  thrpt  100   4.460 ± 0.141  ops/us
QueueThroughputBackoffNone.tpt:offersFailed       131072               MpmcArrayQueue  thrpt  100   0.029 ± 0.031  ops/us
[...]

PR #273, opened 2019-11-16
Fixes #272 BurstCost aren't handling multi-consumers correctly

### LINKED ISSUE #272: BurstCost aren't handling multi-consumers correctly
MpqBurstCost and QueueBurstCost aren't handling correctly multi-consumer scenario: see https://github.com/netty/netty/pull/9078#issuecomment-485066059 for further explanations.

franz1981:
This is fixing MpqBurstCost and QueueBurstCost multi-consumer scenario
and will make both to look similar.

### DIFF
--- a/jctools-benchmarks/src/main/java/org/jctools/jmh/latency/MpqBurstCost.java
+++ b/jctools-benchmarks/src/main/java/org/jctools/jmh/latency/MpqBurstCost.java
@@ -86,19 +86,22 @@ public class MpqBurstCost
     }
 
     @Setup(Level.Iteration)
-    public void startConsumers()
+    public void startConsumers() throws InterruptedException
     {
         stopped = new CountDownLatch(consumerCount);
+        final CountDownLatch started = new CountDownLatch(consumerCount);
         final int consumerCount = this.consumerCount;
         for (int i = 0; i < consumerCount; i++)
         {
             consumers[i].isRunning = true;
             consumers[i].stopped = stopped;
+            consumers[i].started = started;
         }
         for (int i = 0; i < consumerCount; i++)
         {
             consumerExecutor.execute(consumers[i]);
         }
+        started.await();
     }
 
     @TearDown(Level.Iteration)
@@ -241,6 +244,7 @@ public class MpqBurstCost
         MessagePassingQueue<Event> q;
         volatile boolean isRunning = true;
         CountDownLatch stopped;
+        CountDownLatch started;
     }
 
     static class Consumer extends ConsumerFields implements Runnable
@@ -259,8 +263,10 @@ public class MpqBurstCost
         public void run()
         {
             final CountDownLatch stopped = this.stopped;
+            final CountDownLatch started = this.started;
             final int consumerId = this.consumerId;
             final MessagePassingQueue<Event> q = this.q;
+            started.countDown();
             while (isRunning)
             {
                 consume(q, consumerId);
--- a/jctools-benchmarks/src/main/java/org/jctools/jmh/latency/MpqBurstCost.java
+++ b/jctools-benchmarks/src/main/java/org/jctools/jmh/latency/MpqBurstCost.java
@@ -15,10 +15,14 @@ package org.jctools.jmh.latency;
 
 import org.jctools.queues.MessagePassingQueue;
 import org.jctools.queues.MessagePassingQueueByTypeFactory;
+import org.jctools.util.UnsafeAccess;
 import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
 
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 @State(Scope.Benchmark)
 @BenchmarkMode(Mode.AverageTime)
@@ -28,129 +32,191 @@ import java.util.concurrent.atomic.AtomicBoolean;
 @SuppressWarnings("serial")
 public class MpqBurstCost
 {
-    static final Go GO = new Go();
+    private static final long DELAY_PRODUCER = Long.getLong("delay.p", 0L);
+    private static final long DELAY_CONSUMER = Long.getLong("delay.c", 0L);
     @Param( {"SpscArrayQueue", "MpscArrayQueue", "SpmcArrayQueue", "MpmcArrayQueue"})
     String qType;
     @Param( {"100"})
     int burstSize;
     @Param("1")
     int consumerCount;
-    MessagePassingQueue<AbstractEvent> q;
-    private Thread[] consumerThreads;
+    @Param("true")
+    boolean warmup;
+    @Param(value = {"132000"})
+    int qCapacity;
+    MessagePassingQueue<Event> q;
+    private ExecutorService consumerExecutor;
     private Consumer consumer;
 
     @Setup(Level.Trial)
     public void setupQueueAndConsumers()
     {
-        q = MessagePassingQueueByTypeFactory.createQueue(qType, 128);
+        if (warmup)
+        {
+            q = MessagePassingQueueByTypeFactory.createQueue(qType, 128);
 
-        // stretch the queue to the limit, working through resizing and full
-        for (int i = 0; i < 128 + 100; i++)
-        {
-            q.offer(GO);
+            final Event event = new Event();
+
+            // stretch the queue to the limit, working through resizing and full
+            for (int i = 0; i < 128 + 100; i++)
+            {
+                q.offer(event);
+            }
+            for (int i = 0; i < 128 + 100; i++)
+            {
+                q.poll();
+            }
+            // make sure the important common case is exercised
+            for (int i = 0; i < 20000; i++)
+            {
+                q.offer(event);
+                q.poll();
+            }
         }
-        for (int i = 0; i < 128 + 100; i++)
-        {
-            q.poll();
-        }
-        // make sure the important common case is exercised
-        for (int i = 0; i < 20000; i++)
-        {
-            q.offer(GO);
-            q.poll();
-        }
-        q = MessagePassingQueueByTypeFactory.createQueue(qType, 128 * 1024);
-        consumer = new Consumer(q);
-        consumerThreads = new Thread[consumerCount];
+        q = MessagePassingQueueByTypeFactory.createQueue(qType, qCapacity);
+        consumer = new Consumer(q, consumerCount > 1);
+        consumerExecutor = Executors.newFixedThreadPool(consumerCount);
+    }
+
+    @Setup(Level.Iteration)
+    public void startConsumers()
+    {
+        consumer.isRunning = true;
+        consumer.stopped = new CountDownLatch(consumerCount);
         for (int i = 0; i < consumerCount; i++)
         {
-            consumerThreads[i] = new Thread(consumer);
-            consumerThreads[i].start();
+            consumerExecutor.execute(consumer);
         }
+
+    }
+
+    @TearDown(Level.Iteration)
+    public void stopConsumers() throws InterruptedException
+    {
+        consumer.isRunning = false;
+        consumer.stopped.await();
+    }
+
+    @TearDown(Level.Trial)
+    public void stopExecutor() throws InterruptedException
+    {
+        consumerExecutor.shutdown();
     }
 
     @Benchmark
-    public void burstCost(Stop stop)
+    @CompilerControl(CompilerControl.Mode.DONT_INLINE)
+    public void burstCost(Event event)
     {
         final int burst = burstSize;
-        final MessagePassingQueue<AbstractEvent> q = this.q;
-        final Go go = GO;
-        stop.lazySet(false);
-        for (int i = 0; i < burst - 1; i++)
+        final MessagePassingQueue<Event> q = this.q;
+        event.reset();
+        sendBurst(q, event, burst);
+        event.waitFor(burst);
+    }
+
+    private static void sendBurst(MessagePassingQueue<Event> q, Event event, int burst)
+    {
+        for (int i = 0; i < burst; i++)
         {
-            while (!q.offer(go))
+            while (!q.offer(event))
             {
                 ;
             }
-        }
-
-        while (!q.offer(stop))
-        {
-            ;
-        }
-
-        while (!stop.get())
-        {
-            ;
+            if (DELAY_PRODUCER != 0)
+            {
+                Blackhole.consumeCPU(DELAY_PRODUCER);
+            }
         }
     }
 
-    @TearDown
-    public void killConsumer() throws InterruptedException
-    {
-        consumer.isRunning = false;
-        for (int i = 0; i < consumerCount; i++)
-        {
-            consumerThreads[i].join();
-        }
-    }
-
-    abstract static class AbstractEvent extends AtomicBoolean
-    {
-        abstract void handle();
-    }
-
-    static class Go extends AbstractEvent
-    {
-        @Override
-        void handle()
-        {
-            // do nothing
-        }
-    }
-
+    //do not extends AtomicLong here: JMH will take care of padding fields for us
     @State(Scope.Thread)
-    public static class Stop extends AbstractEvent
+    public static class Event
     {
-        @Override
-        void handle()
+        private static final long handledOffset = UnsafeAccess.fieldOffset(Event.class, "handled");
+        private long handled;
+
+        void reset()
         {
-            lazySet(true);
+            UnsafeAccess.UNSAFE.putOrderedLong(this, handledOffset, 0);
+        }
+
+        void waitFor(int value)
+        {
+            while (UnsafeAccess.UNSAFE.getLongVolatile(this, handledOffset) != value)
+            {
+
+            }
+        }
+
+        void sharedHandle()
+        {
+            UnsafeAccess.UNSAFE.getAndAddLong(this, handledOffset, 1);
+        }
+
+        void exclusiveHandle()
+        {
+            UnsafeAccess.UNSAFE.putOrderedLong(this, handledOffset, handled + 1);
         }
     }
 
-    static class Consumer implements Runnable
+    static class ConsumerPad
     {
-        final MessagePassingQueue<AbstractEvent> q;
-        volatile boolean isRunning = true;
+        public long p40, p41, p42, p43, p44, p45, p46;
+        public long p30, p31, p32, p33, p34, p35, p36, p37;
+    }
 
-        public Consumer(MessagePassingQueue<AbstractEvent> q)
+    static class ConsumerFields extends ConsumerPad
+    {
+        MessagePassingQueue<Event> q;
+        volatile boolean isRunning = true;
+        CountDownLatch stopped;
+    }
+
+    static class Consumer extends ConsumerFields implements Runnable
+    {
+        public long p40, p41, p42, p43, p44, p45, p46;
+        public long p30, p31, p32, p33, p34, p35, p36, p37;
+        private final boolean shared;
+
+        public Consumer(MessagePassingQueue<Event> q, boolean shared)
         {
             this.q = q;
+            this.shared = shared;
         }
 
         @Override
         public void run()
         {
-            final MessagePassingQueue<AbstractEvent> q = this.q;
+            final CountDownLatch stopped = this.stopped;
+            final MessagePassingQueue<Event> q = this.q;
+            final boolean shared = this.shared;
             while (isRunning)
             {
-                AbstractEvent e = null;
-                if ((e = q.relaxedPoll()) == null)
-                {
-                    continue;
-                }
-                e.handle();
+                consume(q, shared);
+            }
+            stopped.countDown();
+        }
+
+        @CompilerControl(CompilerControl.Mode.DONT_INLINE)
+        private void consume(MessagePassingQueue<Event> q, boolean shared)
+        {
+            Event e = null;
+            if ((e = q.relaxedPoll()) == null)
+            {
+                return;
+            }
+            if (DELAY_CONSUMER != 0)
+            {
+                Blackhole.consumeCPU(DELAY_CONSUMER);
+            }
+            if (shared)
+            {
+                e.sharedHandle();
+            }
+            else
+            {
+                e.exclusiveHandle();
             }
         }
 
--- a/jctools-benchmarks/src/main/java/org/jctools/jmh/latency/MpqBurstCost.java
+++ b/jctools-benchmarks/src/main/java/org/jctools/jmh/latency/MpqBurstCost.java
@@ -19,6 +19,7 @@ import org.jctools.util.UnsafeAccess;
 import org.openjdk.jmh.annotations.*;
 import org.openjdk.jmh.infra.Blackhole;
 
+import java.util.Arrays;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -46,7 +47,8 @@ public class MpqBurstCost
     int qCapacity;
     MessagePassingQueue<Event> q;
     private ExecutorService consumerExecutor;
-    private Consumer consumer;
+    private Consumer[] consumers;
+    private CountDownLatch stopped;
 
     @Setup(Level.Trial)
     public void setupQueueAndConsumers()
@@ -74,31 +76,43 @@ public class MpqBurstCost
             }
         }
         q = MessagePassingQueueByTypeFactory.createQueue(qType, qCapacity);
-        consumer = new Consumer(q, consumerCount > 1);
+        consumers = new Consumer[consumerCount];
+        for (int i = 0; i < consumerCount; i++)
+        {
+            consumers[i] = new Consumer(q, i);
+        }
         consumerExecutor = Executors.newFixedThreadPool(consumerCount);
     }
 
     @Setup(Level.Iteration)
     public void startConsumers()
     {
-        consumer.isRunning = true;
-        consumer.stopped = new CountDownLatch(consumerCount);
+        stopped = new CountDownLatch(consumerCount);
+        final int consumerCount = this.consumerCount;
         for (int i = 0; i < consumerCount; i++)
         {
-            consumerExecutor.execute(consumer);
+            consumers[i].isRunning = true;
+            consumers[i].stopped = stopped;
         }
-
+        for (int i = 0; i < consumerCount; i++)
+        {
+            consumerExecutor.execute(consumers[i]);
+        }
     }
 
     @TearDown(Level.Iteration)
     public void stopConsumers() throws InterruptedException
     {
-        consumer.isRunning = false;
-        consumer.stopped.await();
+        final int consumerCount = this.consumerCount;
+        for (int i = 0; i < consumerCount; i++)
+        {
+            consumers[i].isRunning = false;
+        }
+        stopped.await();
     }
 
     @TearDown(Level.Trial)
-    public void stopExecutor() throws InterruptedException
+    public void stopExecutor()
     {
         consumerExecutor.shutdown();
     }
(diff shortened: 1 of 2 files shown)


franz1981:
Now I'm running the Queue-one to check again the results on http://psy-lob-saw.blogspot.com/2015/01/mpmc-multi-multi-queue-vs-clq.html ;)
This fix is related to the mpmc xadd q too: I would like to run it vs https://github.com/JCTools/JCTools/pulls to compare with MpmcArrayQueue with multiple consumers

franz1981:
I could have used a different approach to track event handlings ie by not using an AtomicLong shared between consumers, but a long[] per-producer (thread) where each consumer is responsible of a specific index. It would make the `waitFor` slower, but won't incour into the `xadd` cost of `AtomicLong` that would make handling contended. Probably I will add a commit with this approach too.

franz1981:
With the first version of the benchmark (by using xadd when consumers > 1) 
with `burstSize = 100` I'm getting:
```
---------
1P-1C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt      Score     Error  Units
QueueBurstCost.burstCost          100                1       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  11318.278 ±  91.259  ns/op
QueueBurstCost.burstCost          100                1       132000        ConcurrentLinkedQueue      true  avgt   40  17116.521 ± 480.471  ns/op
QueueBurstCost.burstCost          100                1       132000               MpmcArrayQueue      true  avgt   40   7148.785 ± 236.638  ns/op
---------
2P-2C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt      Score     Error  Units
QueueBurstCost.burstCost          100                2       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  23409.568 ± 355.988  ns/op
QueueBurstCost.burstCost          100                2       132000        ConcurrentLinkedQueue      true  avgt   40  40930.614 ± 876.973  ns/op
QueueBurstCost.burstCost          100                2       132000               MpmcArrayQueue      true  avgt   40  26813.796 ± 670.512  ns/op
---------
3P-3C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt      Score      Error  Units
QueueBurstCost.burstCost          100                3       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  37742.824 ±  486.882  ns/op
QueueBurstCost.burstCost          100                3       132000        ConcurrentLinkedQueue      true  avgt   40  67822.700 ± 3790.893  ns/op
QueueBurstCost.burstCost          100                3       132000               MpmcArrayQueue      true  avgt   40  49307.053 ± 1144.943  ns/op
---------
4P-4C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score      Error  Units
QueueBurstCost.burstCost          100                4       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40   64889.469 ± 2997.470  ns/op
QueueBurstCost.burstCost          100                4       132000        ConcurrentLinkedQueue      true  avgt   40  112583.181 ± 8774.015  ns/op
QueueBurstCost.burstCost          100                4       132000               MpmcArrayQueue      true  avgt   40   67236.589 ± 2185.796  ns/op
---------
5P-5C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score      Error  Units
QueueBurstCost.burstCost          100                5       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  102090.251 ± 1972.969  ns/op
QueueBurstCost.burstCost          100                5       132000        ConcurrentLinkedQueue      true  avgt   40  183607.705 ± 2645.447  ns/op
QueueBurstCost.burstCost          100                5       132000               MpmcArrayQueue      true  avgt   40   93294.382 ± 3663.733  ns/op
---------
6P-6C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score      Error  Units
QueueBurstCost.burstCost          100                6       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  137735.752 ± 2284.493  ns/op
QueueBurstCost.burstCost          100                6       132000        ConcurrentLinkedQueue      true  avgt   40  248958.667 ± 3797.790  ns/op
QueueBurstCost.burstCost          100                6       132000               MpmcArrayQueue      true  avgt   40  122083.077 ± 5117.835  ns/op
```
I will stress just a couple of asymmetric configurations to make sure that the xadd q is delivering its promises:
```
---------
10P-2C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score      Error  Units
QueueBurstCost.burstCost          100                2       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  125105.547 ± 3191.048  ns/op
QueueBurstCost.burstCost          100                2       132000        ConcurrentLinkedQueue      true  avgt   40  207516.138 ± 3626.670  ns/op
QueueBurstCost.burstCost          100                2       132000               MpmcArrayQueue      true  avgt   40  278119.574 ± 9714.185  ns/op
---------
8P-4C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score       Error  Units
QueueBurstCost.burstCost          100                4       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  130007.279 ±  5887.400  ns/op
QueueBurstCost.burstCost          100                4       132000        ConcurrentLinkedQueue      true  avgt   40  203753.285 ± 14697.797  ns/op
QueueBurstCost.burstCost          100                4       132000               MpmcArrayQueue      true  avgt   40  204236.319 ± 18750.423  ns/op
```

franz1981:
I've added a new commit, trying to save using xadd consumer-side, by using separate counters per-consumer, but obviously it would hit false-sharing between counters: this could be saved by using 2*CACHE_LINE pads between each counter to avoid consumers invalidating each-others cache-lines while handling producers tasks.
I'm reporting here the same results of https://github.com/JCTools/JCTools/pull/273#issuecomment-554620865 to see if there is any improvement on the results.
```
---------
1P-1C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt      Score     Error  Units
QueueBurstCost.burstCost          100                1       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  11583.685 ±  58.408  ns/op
QueueBurstCost.burstCost          100                1       132000        ConcurrentLinkedQueue      true  avgt   40  17266.134 ± 320.246  ns/op
QueueBurstCost.burstCost          100                1       132000               MpmcArrayQueue      true  avgt   40   6979.031 ± 125.073  ns/op
---------
2P-2C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt      Score      Error  Units
QueueBurstCost.burstCost          100                2       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  23675.018 ±  499.502  ns/op
QueueBurstCost.burstCost          100                2       132000        ConcurrentLinkedQueue      true  avgt   40  46451.762 ± 3138.152  ns/op
QueueBurstCost.burstCost          100                2       132000               MpmcArrayQueue      true  avgt   40  26189.720 ± 1318.632  ns/op
---------
3P-3C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt      Score      Error  Units
QueueBurstCost.burstCost          100                3       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  43946.182 ± 2227.768  ns/op
QueueBurstCost.burstCost          100                3       132000        ConcurrentLinkedQueue      true  avgt   40  87120.460 ± 3194.932  ns/op
QueueBurstCost.burstCost          100                3       132000               MpmcArrayQueue      true  avgt   40  45105.417 ±  929.084  ns/op
---------
4P-4C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score      Error  Units
QueueBurstCost.burstCost          100                4       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40   70045.186 ± 2241.860  ns/op
QueueBurstCost.burstCost          100                4       132000        ConcurrentLinkedQueue      true  avgt   40  137313.832 ± 1815.112  ns/op
QueueBurstCost.burstCost          100                4       132000               MpmcArrayQueue      true  avgt   40   68284.462 ± 1983.369  ns/op
---------
5P-5C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score      Error  Units
QueueBurstCost.burstCost          100                5       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  100252.444 ± 3413.939  ns/op
QueueBurstCost.burstCost          100                5       132000        ConcurrentLinkedQueue      true  avgt   40  191076.869 ± 2130.139  ns/op
QueueBurstCost.burstCost          100                5       132000               MpmcArrayQueue      true  avgt   40   95408.004 ± 4680.241  ns/op
---------
6P-6C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score      Error  Units
QueueBurstCost.burstCost          100                6       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  134966.416 ± 1555.916  ns/op
QueueBurstCost.burstCost          100                6       132000        ConcurrentLinkedQueue      true  avgt   40  249424.726 ± 2469.640  ns/op
QueueBurstCost.burstCost          100                6       132000               MpmcArrayQueue      true  avgt   40  127035.803 ± 5283.807  ns/op
```
With asymmetric configurations:
```
---------
11P-1C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score       Error  Units
QueueBurstCost.burstCost          100                1       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40   67192.319 ±  4189.873  ns/op
QueueBurstCost.burstCost          100                1       132000        ConcurrentLinkedQueue      true  avgt   40  285255.400 ± 19155.638  ns/op
QueueBurstCost.burstCost          100                1       132000               MpmcArrayQueue      true  avgt   40  315123.221 ±  4491.199  ns/op
---------
10P-2C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score       Error  Units
QueueBurstCost.burstCost          100                2       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  117949.236 ±  4571.083  ns/op
QueueBurstCost.burstCost          100                2       132000        ConcurrentLinkedQueue      true  avgt   40  267017.522 ±  8036.012  ns/op
QueueBurstCost.burstCost          100                2       132000               MpmcArrayQueue      true  avgt   40  284326.729 ± 10299.422  ns/op
---------
8P-4C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score       Error  Units
QueueBurstCost.burstCost          100                4       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  137988.606 ±  4540.951  ns/op
QueueBurstCost.burstCost          100                4       132000        ConcurrentLinkedQueue      true  avgt   40  283946.006 ±  3914.199  ns/op
QueueBurstCost.burstCost          100                4       132000               MpmcArrayQueue      true  avgt   40  222860.950 ± 12340.679  ns/op
```
The "wild" machine cases too:
```
---------
22P-22C:
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt        Score        Error  Units
QueueBurstCost.burstCost          100               22       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40   489316.718 ±  41842.624  ns/op
QueueBurstCost.burstCost          100               22       132000        ConcurrentLinkedQueue      true  avgt   40  1901868.898 ±  53407.107  ns/op
QueueBurstCost.burstCost          100               22       132000               MpmcArrayQueue      true  avgt   40   971550.149 ± 150254.715  ns/op
---------
8P-8C
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score       Error  Units
QueueBurstCost.burstCost          100                8       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  190692.631 ±  7632.108  ns/op
QueueBurstCost.burstCost          100                8       132000        ConcurrentLinkedQueue      true  avgt   40  805924.222 ± 16923.904  ns/op
QueueBurstCost.burstCost          100                8       132000               MpmcArrayQueue      true  avgt   40  279469.248 ± 12048.981  ns/op
---------
6P-6C
---------
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt       Score       Error  Units
QueueBurstCost.burstCost          100                6       132000  MpmcUnboundedXaddArrayQueue      true  avgt   40  126767.559 ±  5528.759  ns/op
QueueBurstCost.burstCost          100                6       132000        ConcurrentLinkedQueue      true  avgt   40  466296.406 ± 51349.549  ns/op
QueueBurstCost.burstCost          100                6       132000               MpmcArrayQueue      true  avgt   40  200847.577 ± 15646.834  ns/op
```
Some conclusions from the asymmetric tests: I don't think `MpmcArrayQueue` has something to envy to CLQ although is getting worst results then it, but that given the consumer count is lowered, the chance that consumers are able to getting empty q is higher, invalidating the producer sequence in order to be sure of emptyness; probably that's the reason why it seems slower then CLQ.

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
```diff
@@ -44,22 +44,25 @@
     @Param("true")
     boolean warmup;
     @Param(value = {"132000"})
-    String qCapacity;
-    Queue<AbstractEvent> q;
+    int qCapacity;
```
nitsanw: qCapacity is a string to allow the <chunk,size> format conveniently from the command line.

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
```diff
@@ -68,102 +71,145 @@ public void setupQueue()
...
-                ;
+                ELEMENT_SHIFT = 3;
             }
-            if (DELAY_PRODUCER != 0)
+            else
             {
-                Blackhole.consumeCPU(DELAY_PRODUCER);
+                throw new IllegalStateException("Unexpected long[] element size");
             }
+            // Including the buffer pad in the array base offset
```
nitsanw: What buffer pad?

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
```diff
@@ -68,102 +71,145 @@ public void setupQueue()
...
+            }
+            while (true);
+        }
+
+        void handle(int consumerId)
+        {
+            final long[] values = this.handledCount;
+            final long offset = calcSequenceOffset(consumerId);
+            final long value = UnsafeAccess.UNSAFE.getLong(values, offset);
+            UnsafeAccess.UNSAFE.putOrderedLong(values, offset, value + 1);
```
nitsanw: This introduces significant false sharing on the consumer counters

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
```diff
@@ -68,102 +71,145 @@ public void setupQueue()
...
 
-    static class Go extends AbstractEvent
-    {
-        @Override
-        void handle()
+        @Setup
+        public void init(QueueBurstCost state)
         {
-            // do nothing
+            handledCount = new long[state.consumerCount];
```
nitsanw: I think it would be far less complex to put the counter on Consumer, which already has padding. The `waitFor` then looks at the counters on the consumers. You can reduce the reader induced coherence overhead by spin waiting on the final event (as before) and then transition to consumer scanning. Further you can batch update the `handled` counter on the Consumer only when the queue is observed empty or relaxedPoll returns null.

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
franz1981: I have noted that one of the bench was hard-coding it: I can make both `String` if is better, ok?

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
franz1981: Ups, copied from:  https://github.com/JCTools/JCTools/blob/172a2f87a2b381cb7bee481df6b72a4abc07a261/jctools-core/src/main/java/org/jctools/queues/ConcurrentSequencedCircularArrayQueue.java#L40

Let me remove it (if the long[] approach will survive)

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
franz1981: > I think it would be far less complex to put the counter on Consumer, which already has padding

That's a nice idea, but the counter is not just per-consumer, need to be per-producer (thread) as well

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
franz1981: That's correct (as noted by https://github.com/JCTools/JCTools/pull/273#issuecomment-554743228), let me quote myself:
> this could be saved by using 2*CACHE_LINE pads between each counter to avoid consumers invalidating each-others cache-lines while handling producers tasks.

it can works?

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
nitsanw: hard coding it?

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
nitsanw: "That's a nice idea, but the counter is not just per-consumer, need to be per-producer (thread) as well."
Fair point, in which case, yes you have to pad the array based counter.

"This would work if the last offered event will be the last to be handled, but with multi-consumer is not ensured anymore"
You can't be sure, but it's fair to assume that there's no need to check the counters before that one in handled.

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
franz1981: Yep: https://github.com/JCTools/JCTools/pull/273/files#diff-01928db23694a1a5eea730064785eb33L62

There was no means to change it in the original version: that's why I've tried to make both the bench to look similar

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
franz1981: > it's fair to assume that there's no need to check the counters before that one in handled

I don't understand this: maybe with "before" I mean something different.
My understanding of "before" is the index/id of consumer on `Event::handledCount`: if the last (sbmitted) event is being handled, only with a single consumer I can trust that the previous events would be hanlded by it in order, saving the needto poll for their completion, but the current algorithm won't do that anyway (because `Event::handledCount::length == consumerCount == 1`, you ends up always polling).
Help me to understand what do you mean, sorry for being slow in this :)

nitsanw: "I can trust that the previous events would be handled with offering order just in the single consumer case, saving to poll for theirs completion."
I'm not saying you should trust this as the only indicator, just that it's a good heuristic to reduce the overall noise.

franz1981: Ok, so I suppose that you mean the `before` as "previous handled events by that same consume": in this case, yes I can handle it like that and see how it goes!

OT:
Do you have had any chance to see the results I've posted? They should ideally replace the ones on http://psy-lob-saw.blogspot.com/2015/01/mpmc-multi-multi-queue-vs-clq.html given that now the mc case should be handled correctly. I'm especially curious about your opinion/experience with asymmetric bench results too.

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
franz1981: I've addressed it in the last commit: let me resolve this

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
franz1981: Given that we're going to deal with false sharing through padding, I'm going to resolve this

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
franz1981: I've handled false-sharing in the last commit

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/QueueBurstCost.java
franz1981: I haven't yet addressed the last change (ie to use last event handling), if you have any concern please send a snipped with the idea you propose to speedup the fix,
I need yet to understand the value of 
```
        @CompilerControl(CompilerControl.Mode.DONT_INLINE)
        private void consume(MessagePassingQueue<Event> q, int consumerId)
```
Why isn't good to be inlined? 
By preventing inlining It would artificially slow don't too fast consumers (that has an high chance of finding the q emtpy) due to increased cost of `consume` when no events are present, thus producing noise.

nitsanw: WRT benchmark results, I would appreciate it if you made the effort to format those into something that is more easily comparable.

nitsanw: WRT the `consume` method inlining, I can't remember. There's good reasons to avoid inlining in infinite loops to avoid OSR compilations, but I can't remember what was the reasons/testing here.

franz1981: Configuration is the same as https://github.com/franz1981/JCTools-wiki/blob/xadd_docs/MpscUnboundedXaddArrayQueue.md#lies-damned-lies-and-benchmarks
*Symmetric tests*
1P-1C:
image
2P-2C:
image
3P-3C:
image
4P-4C:
image
5P-5C:
image
6P-6C:
image

franz1981: 8P-4C:
image
10P-2C:
image
11P-1C:
image

franz1981: @nitsanw Let me know any insights about them and how you think is performing xadd q in your opinion @ https://github.com/JCTools/JCTools/pull/269

franz1981:
Configuration is the same as https://github.com/franz1981/JCTools-wiki/blob/xadd_docs/MpscUnboundedXaddArrayQueue.md#lies-damned-lies-and-benchmarks and test has been run using:
```
$ for i in 1 2 3 4 5 6; do java -jar microbenchmarks.jar org.jctools.jmh.latency.QueueBurstCost.* -f 10 -t $i -p consumerCount=$i -p burstSize=100 -p qType=MpmcUnboundedXaddArrayQueue,ConcurrentLinkedQueue,MpmcArrayQueue -rf json -rff $i.json >> burst.$i.txt; done
```
*Symmetric tests*
1P-1C:
image
2P-2C:
image
3P-3C:
image
4P-4C:
image
5P-5C:
image
6P-6C:
image

franz1981:
@nitsanw If you want me to run on the same machine some specific tests, do not esitate to ask bud :)

nitsanw:
@franz1981  These are interesting. How do they compare to the prev.(before the fix) results?

franz1981:
@nitsanw Do you mean the results of the previous version of the benchmark (before *this* PR) or before the fix (on https://github.com/JCTools/JCTools/pull/269) of the xadd queue?

FYI the benchmark results collected in this PR are with https://github.com/JCTools/JCTools/pull/269 applied.

franz1981:
@nitsanw 
I'm re-running the same benchmarks of https://github.com/JCTools/JCTools/pull/273#issuecomment-555237103 on `master`, but with #269 applied.
If the results will show that the results are sounds (for this PR) I will perform the same vs `master` without #269 applied to verify that #269 seems a good change (throughput tests already confirmed it).

nitsanw:
#269 is a different story, I'm interested in the benchmark results before/after the change to the benchmarks. In particular we would expect all the single consumer scenarios to produce the same results.

nitsanw:
To be clear, I'm not at all interested here in the comparison between different queue implementations so much as in the effect of the benchmark change. So the visualisation, while nice, still requires me to scroll up and down and look for the b/a picture.

franz1981:
> In particular we would expect all the single consumer scenarios to produce the same results.

That's what the numbers says, assuming some some error due to my use of `taskset`: I'm not using `taskset` to let the benchs run on just 2 different cores, but always on 12 different cores belonging the same numa node, so both benchs can see some error due to context switches/core migrations.
I can re-run both benchs in the 1P-1C case while pinning it just on 2 cores, if you think it worths.

> So the visualisation, while nice, still requires me to scroll up and down and look for the b/a picture.

Given that NP-NC with N > 1 are totally not-comparable (the original code with multi-consumer is just broken IIRC) I can manage to get a single run of both benchs with `MpmcArrayQueue` alone and sees what come out.
<joking-mode>
Please, hold your breath and save those expensive scrolling movements to compare charts: we're not athletes!! :D
</joking-mode>

franz1981:
>  I'm not at all interested here in the comparison between different queue implementations
I can open a separate issue or write on the PR re the mpmc xadd fix if you prefer, but I need some magic help from you there, bud :)

franz1981:
In addition to the results of 1P-1C on https://github.com/JCTools/JCTools/pull/273#issuecomment-557015111, I'm adding these ones too:
2P-1C:
image

4P-1C:
image

8P-1C:
image

11P-1C:
image

TBH I was expecting the same trend of the 1P-1C results ie the new PR to slow down a bit the task execution, but seems not the case.

jctools-benchmarks/src/main/java/org/jctools/jmh/latency/MpqBurstCost.java
```diff
@@ -28,130 +34,253 @@
...
+        for (int i = 0; i < consumerCount; i++)
         {
-            q.offer(GO);
-            q.poll();
+            consumerExecutor.execute(consumers[i]);
         }
-        q = MessagePassingQueueByTypeFactory.createQueue(qType, 128 * 1024);
-        consumer = new Consumer(q);
-        consumerThreads = new Thread[consumerCount];
+    }
```
nitsanw: should probably have a count down latch to make the init wait for the consumers to "start" otherwise we will have some noise from consumer scheduling delay.

nitsanw:
"TBH I was expecting the same trend of the 1P-1C results ie the new PR to slow down a bit the task execution, but rather it seems the opposite" - A slower consumer means less contention. This is a case where consumers are much faster than producers (no contention) so the queue has no slack.

franz1981:
> A slower consumer means less contention

That makes sense to me, but how to make it more evident on the benchmark results? 
With JMH queue throughput benchs we've the failed polls stats, while here we have to profile/perfasm to understand what's going on. Getting the failed poll stats from the beginning till the end of a burst could be helpful although I've no (yet) any idea how to collect these info here.

Returning on topic: this PR's `handle` seems more heavyweight then `master`; as you've suggested it can be made lighter for the single-consumer case, although at the cost a bimorphic call on consumer. But I see 2 benefits of this PR as it is:

1. single/multi-consumer tests behave the same and can be compared (ie `handle` cost is the same for both)
2. the error bars on 1P-1C shows that the difference from master is not big

I'm going to address https://github.com/JCTools/JCTools/pull/273#discussion_r349990966 and wil wait your opinion on how to proceed for this.

nitsanw:
I think with the start latch it's good to merge. Sorry for the long review process :-)

franz1981:
@nitsanw No worries bud; that's the base where we collect numbers to understand what's going on and is sufficiently delicate that need some care while modifying it :)
Given me few minutes that I write the latch change and will push the commit: do you want me to squash all the commits into 1?

franz1981:
I've already pushed the commit with the started latch change: the only downside of it is that one of the consumer(s) will pay the cost of awaking it and unblock `startConsumers`, but it shouldn't be relevant, given that the bench iteration won't start until `startConsumers` will end.

PR #276, opened 2019-12-12
wait-free relaxedPoll for mpmc xadd q

franz1981:
it provides a full-fat wait-free `relaxedPoll` for the mpmc xadd q.
Soon I will publish some results to check how/if it improves if compared with `poll`.
I'm not sure I will provide `relaxedPeek` unless strictly necessary.

My concern is that given the many indirections of this queue, a relaxedPoll won't give any perf benefit if not to be sure that the operation completes in a finite number of steps (as a wait-free one is expecting to do).

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
@@ -459,7 +459,7 @@ public class MpmcUnboundedXaddArrayQueue<E> extends MpmcUnboundedXaddArrayQueueP
         next.soElement(consumerOffset, null);
         next.spPrev(null);
         //save from nepotism
-        consumerBuffer.spNext(null);
+        consumerBuffer.soNext(null);
         if (consumerBuffer.isPooled())
         {
             final boolean offered = freeBuffer.offer(consumerBuffer);
@@ -493,18 +493,21 @@ public class MpmcUnboundedXaddArrayQueue<E> extends MpmcUnboundedXaddArrayQueueP
             firstElementOfNewChunk = consumerOffset == 0 && consumerIndex >= chunkSize;
             if (firstElementOfNewChunk)
             {
-                next = consumerBuffer.lvNext();
-                final long expectedChunkIndex = chunkIndex - 1;
                 //we don't care about < or >, because if:
                 //- consumerBuffer::index < expectedChunkIndex: another consumer has rotated consumerBuffer,
                 //  but not reused (yet, if possible)
                 //- consumerBuffer::index > expectedChunkIndex: another consumer has rotated consumerBuffer,
                 // that has been pooled and reused again
                 //In both cases we have a stale view of the world with a not reliable next value.
+                final long expectedChunkIndex = chunkIndex - 1;
                 if (expectedChunkIndex != consumerBuffer.lvIndex())
                 {
                     continue;
                 }
+                next = consumerBuffer.lvNext();
+                //next could have been modified by another consumer, but:
+                //- if null: it still needs to check q empty + casConsumerIndex
+                //- if !null: it will fail on casConsumerIndex
                 if (next == null)
                 {
                     if (consumerIndex >= pIndex && // test against cached pIndex
@@ -532,6 +535,10 @@ public class MpmcUnboundedXaddArrayQueue<E> extends MpmcUnboundedXaddArrayQueueP
                             //stale view of the world
                             continue;
                         }
+                        //it cover both cases:
+                        //- right chunk, awaiting element to be set
+                        //- old chunk, awaiting rotation
+                        //It allows to fail fast if the q is empty after the first element on the new chunk.
                         if (consumerIndex >= pIndex && // test against cached pIndex
                             consumerIndex == (pIndex = lvProducerIndex()))
                         { // update pIndex if we must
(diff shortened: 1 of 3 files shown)


jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
```diff
@@ -658,7 +658,97 @@ public boolean relaxedOffer(E e)
...
+            if (pooled)
+            {
+                if (next.lvSequence(consumerOffset) != chunkIndex)
+                {
+                    return null;
+                }
+            }
+            else
+            {
+                e = next.lvElement(consumerOffset);
```
franz1981: @nitsanw We don't know here if `next::index == chunkIndex`, but we don't care as long as `e !=null` and we succeed to increment `consumerIndex`

franz1981:
@nitsanw By looking at the code I see that a slow consumer could `lvNext` while another would `spNext(null)` on the same `consumerBuffer`: it shouldn't be a big deal given that `casConsumerIndex` will protect the former to perform interesting logic on `consumerBuffer`, but for atomicity/correctness would be better to turn the `spNext` into a `soNext`, wdyt?

It won't save navigating `consumerBuffer::next` first element if `next != null', but will make sure the `next` read will perform atomically.

franz1981:
@nitsanw I've added a new commit to discuss a different way to check if a chunk is correct. 
TLDR 
In the original version on `rotateConsumerBuffer`:
```java
consumerBuffer.soIndex(AtomicChunk.NIL_CHUNK_INDEX);
//...
consumerBuffer.spNext(null);
//..
```
While on `poll` of `firstElementOfNewChunk`:
```java
next = consumerBuffer.lvNext();
final long expectedChunkIndex = chunkIndex - 1;
if (expectedChunkIndex != consumerBuffer.lvIndex())
{
    continue;
}
```
In my original intent the order of checks was to be sure that the `next` value was still valid to be used (ie `consumerBuffer` wasn't marked as a *NIL* chunk) failing fast without fetching the element on next and check for q emptyness.
In later versions of `rotateConsumerBuffer` I've removed `consumerBuffer.soIndex(AtomicChunk.NIL_CHUNK_INDEX)`, hence I don't see a reason now to maintain the same order of checks on `poll`, given that:

1. `casConsumerIndex` already protect racing rotating consumers which `consumerBuffer::next` is `null` because of the winning rotating consumer
2. it would optimize the common case for a rare event

In the new version on `rotateConsumerBuffer`:
```java
consumerBuffer.soNext(null);
```
While on `poll` of `firstElementOfNewChunk`:
```java
final long expectedChunkIndex = chunkIndex - 1;
if (expectedChunkIndex != consumerBuffer.lvIndex())
{
    continue;
}
next = consumerBuffer.lvNext();
```
I'm now running some benchmark to see if It won't introduce any perf regression, but don't merge it please: I need your opinion on the logic.

franz1981:
I cannot see evident regressions on `QueueBurstCost`, but in the 3P-3C case, I'm running it again to be sure of it.

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
```diff
@@ -541,15 +548,18 @@ public E poll()
                         continue;
                     }
                 } else {
-                    e = consumerBuffer.lvElement(consumerOffset);
                     final long index = consumerBuffer.lvIndex();
-                    if (index != chunkIndex || e == null)
+                    if (index != chunkIndex || (e = consumerBuffer.lvElement(consumerOffset)) == null)
```
nitsanw: Prev style is clearer IMO

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
```diff
@@ -532,6 +535,10 @@ public E poll()
                             //stale view of the world
                             continue;
                         }
+                        //it cover both cases:
+                        //- right chunk, awaiting element to be set
+                        //- old chunk, awaiting rotation
+                        //It allows to fail fast if the q is empty after the first element on the new chunk.
```
nitsanw: Inconsistent capitalization, make up your mind :)

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
```diff
@@ -658,7 +668,96 @@ public boolean relaxedOffer(E e)
...
-        return poll();
+        final int chunkMask = this.chunkMask;
+        final int chunkShift = this.chunkShift;
+        final int chunkSize = chunkMask + 1;
+        final long consumerIndex = this.lvConsumerIndex();
+        final AtomicChunk<E> consumerBuffer = this.lvConsumerBuffer();
+        final int consumerOffset = (int) (consumerIndex & chunkMask);
+        final long chunkIndex = consumerIndex >> chunkShift;
+        final boolean firstElementOfNewChunk = consumerOffset == 0 && consumerIndex >= chunkSize;
+        if (firstElementOfNewChunk)
```
nitsanw: Surely this is rare and can be extracted to another method?

franz1981: This is rare, but it always happen, so I wasn't sure to pack it in a separate method: if you think it would improve readability I will do it :+1:

franz1981:
I've spent some more time benchmarking the 3P-3C case both using `taskset` and with more forks.
This PR:
```
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt      Score     Error  Units
QueueBurstCost.burstCost          100                3       132000  MpmcUnboundedXaddArrayQueue      true  avgt  200  45400.100 ± 583.552  ns/op
```

Master:

```
Benchmark                 (burstSize)  (consumerCount)  (qCapacity)                      (qType)  (warmup)  Mode  Cnt      Score     Error  Units
QueueBurstCost.burstCost          100                3       132000  MpmcUnboundedXaddArrayQueue      true  avgt  200  46167.098 ± 592.102  ns/op
```
I can say that there is no regression although the consumer logic checks are changed.

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
franz1981: Fixed ;)

PR #280, opened 2019-12-30
Added new test cases + changed some bits of logic 4 xadd qs

franz1981:


franz1981:
@nitsanw The important bits to look with care are:

- https://github.com/JCTools/JCTools/pull/280/files#diff-a5799178be39d50d28283429c4116b67R355: !needed anymore both for consumers and producers logic
- https://github.com/JCTools/JCTools/pull/280/files#diff-61853935c955f7d3dd424601f34d4ca1R142: usefull to return `null` during a slow rotation on an empty q

franz1981:
@nitsanw Sadly, I haven't had time yet to run any benchmark, tomorrow early morning will try to do that and give another round of code review too

franz1981:
Tomorrow I will give another round to the logic of the low rotation check because I still think it can be improved/make it clearer

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
```diff
@@ -134,6 +139,18 @@ public E poll()
                         break;
                     }
                 }
+                else if (ciChunkOffset != 0 && ccChunkIndex < ciChunkIndex)
```
nitsanw: hmmm, I'm not sure I get the need for this. We're improving the performance of observing empty queue during chunk transition, which seems like an edge case, at the cost of making the code even bulkier...

franz1981: It was a case already handled in the original code to save awaiting a full rotation to complete (after `soConsumerChunk`) before ack q emptiness. 

Is an edge case, but IMO deserves to be treated correctly given that we're not using parity bits on `consumerIndex` to mark rotation states and it would save from additional pauses/spikes while (not) consuming an element. If a queue is empty I wouldn't expect to "wait" to know that, given that `isEmpty` complete in O(1).

nitsanw: that's only on rotation, so an edge case. It's fine to spin on the edge case IMO

jctools-core/src/main/java/org/jctools/queues/MpUnboundedXaddArrayQueue.java
```diff
@@ -321,7 +311,7 @@ private R newOrPooledChunk(R prevChunk, long nextChunkIndex)
     }
 
 
-    R pollNextBuffer(R cChunk, long cIndex)
+    protected final R pollNextBuffer(R cChunk, long cIndex)
```
nitsanw: `final` is fine, but `protected` is public API for anyone extending the queue

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
```diff
@@ -111,27 +116,42 @@ public E poll()
                 // we are looking at the first element of the next chunk
                 // NOTE: The check is different from the single consumer case as the MC case requires checking this
```
nitsanw: change the comments to match the code changes

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
```diff
@@ -111,27 +116,42 @@ public E poll()
...
-                            // strict empty check, this ensures [Queue.poll() == null iff isEmpty()]
-                            return null;
+                            break;
                         }
-                        // we will go ahead with the CAS and have the winning consumer spin for the next buffer
                     }
-                    // not empty: can attempt the cas (and transition to next chunk if successful)
-                    if (casConsumerIndex(cIndex, cIndex + 1))
+                }
+                else if (ccChunkIndex < ciChunkIndex)
```
nitsanw: more indentation -> less readable

jctools-core/src/main/java/org/jctools/queues/MpUnboundedXaddArrayQueue.java
franz1981: So we don't want the subclasses of users to call this, got it, let me change it :+1

nitsanw: public API -> liability

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
franz1981: Let's continue on https://github.com/JCTools/JCTools/pull/280#discussion_r362171799

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
```diff
@@ -133,6 +132,16 @@ public E poll()
                     {
                         break;
                     }
+                } else if (ccChunkIndex < ciChunkIndex) {
```
nitsanw: I still don't see the point of this optimization. If you miss read the chunk you are going to now load the pIndex too. If the queue is actually empty you'll get here soon enough anyhow. If the queue is near empty contended with many consumers then you'll see empty more often, but not worth the extra code IMO.

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
franz1981: I see that `poll' was already prone to some spinning due to the time taken by producers to reach the target chunk and that saving the check won't break the relation with `isEmpty`, but...I'm still not feeling Ok to not "fix" something a potential (rare) issue...
Probably I'm biased on this due to the pain to find a solution to this edge case :(
Hence I will trust your judgment on this.
I'm sure that if Java would support packing this:
```java
if (cIndex >= pIndex && // test against cached pIndex
    cIndex == (pIndex = lvProducerIndex())) // update pIndex if we must
{
    // strict empty check, this ensures [Queue.poll() == null iff isEmpty()]
    return null;
}
```
without creating garbage, this check should have been preserved

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
franz1981: The point is that you cannot answer back for emptiness until a full rotation happen, while you already have that information: more then optimization is a way to use an information we already have without awaiting more the necessary

franz1981: > If you miss read the chunk you are going to now load the pIndex too

But here you're not missing to read the right chunk, but the consumer in charge of rotating `consumerChunk` has won the `casConsumerIndex` (that should point to the second element in the new chunk), but is still awaiting *something* before rotating `consumerChunk` using `soConsumerChunk`.
In this case, the current code will continue to spin, while it shouldn't if the second element in the new chunk is not going to be set.

franz1981: I'm not sure if we're seeing the same issue/case here or I'm misreading (entirely possible) your comment...

PR #281, opened 2020-01-02
mpmc xadd q poll wasn't handling correctly chunkSize = 1 pooled case

franz1981:
QueueSanityTest::testSize with chunkSize = 1 and 1 recycled chunks was
failing with a blocked poll because it wasn't handling ccChunkIndex ==
ciChunkIndex as an isFirstElementOfNextChunk case


jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddChunk.java
```diff
@@ -40,11 +40,25 @@
...
+    void spinForSequence(int index, long e)
+    {
+        assert isPooled();
+        final long[] sequence = this.sequence;
+        final long offset = calcLongElementOffset(index);
+        do
+        {
+
+        }
+        while (lvLongElement(sequence, offset) != e);
```
nitsanw: move the load into the loop, nicer to not have empty loop

PR #282, opened 2020-01-03
mpmc xadd q poll docs + clearing up names

franz1981:


franz1981:
@nitsanw I've tried to move the pooling of `cChunk` after `this.soConsumerChunk(next);`, be can't because `this.soConsumerChunk(next)` mark the end of the single-threaded code and `freeChunksPool` is spsc, so cannot be used by different producers, we would need an mpsc queue to perform that change

PR #283, opened 2020-01-04
xadd qs: cleaning up abstractions

franz1981:


franz1981:
I've noticed that `moveToNextConsumerChunk` is using `soConsumerChunk` for single-consumer case too: TBH the perf difference should be negligible and I cannot say if it worths to address it

PR #285, opened 2020-01-04
Fixes #284 Add support for Thread.onSpinWait() while spinning

### LINKED ISSUE #284: Add support for Thread.onSpinWait() while spinning
It wides the scope of https://github.com/JCTools/JCTools/issues/147, using it on queues by reusing the ideas on https://github.com/giltene/GilExamples/tree/master/SpinWaitTest/src/main/java/org/performancehints to provide support while retaining Java 6 compatibility.

franz1981:
I've opened this for discussion: I'll push different commits to propose places where `onSpinWait` could be welcome, if I'll solve first the compilation error:
```
InternalThreadHints.java:[52,41] method invoked with incorrect number of arguments; expected 0, found 1
```

### DIFF
--- a/jctools-core/src/main/java/org/jctools/util/ThreadHints.java
+++ b/jctools-core/src/main/java/org/jctools/util/ThreadHints.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2016 Gil Tene
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jctools.util;
+
+@InternalAPI
+public final class ThreadHints
+{
+
+    /**
+     * Set this system property to true to disable {@link #onSpinWait()}.
+     */
+    public static final String DISABLE_ON_SPIN_WAIT = "jctools.disable.onSpinWait";
+    private static final boolean okToInvokeInternalThreadHintsCaller;
+
+    static
+    {
+        boolean okToInvokeMHCaller = false;
+        if (!Boolean.getBoolean(DISABLE_ON_SPIN_WAIT))
+        {
+            try
+            {
+                Class.forName("java.lang.invoke.MethodHandle");
+                okToInvokeMHCaller = true;
+            }
+            catch (Exception e)
+            {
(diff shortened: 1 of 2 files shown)


franz1981:
Probably `@PolymorphicSignature` is creating issues while compiling with 1.6, while with 1.7 is fine

franz1981:
Here https://github.com/LMAX-Exchange/disruptor/issues/183#issuecomment-269059148 some more information on how we could bring this in with 1.6, if it worths doing it

richardstartin:
If I understand correctly you want `Thread.onSpinWait` to be used in spin locks whenever it is available while maintaining bytecode compatibility with JDK6. You can achieve this with a shim class and a Multi-Release jar. 

```java
public class ThreadShim {

public void onSpinWait() {
    // do nothing
}

}
```

In another directory e.g. `src/main/java9` you add another implementation.

```java
/** will load this one from META-INF/versions/9 in any JDK > JDK9
public class ThreadShim {

public void onSpinWait() {
    Thread.onSpinWait();
}

}
```

You now need to ensure that the JDK9+ implementation is in a special place in the jar (`META-INF/versions/9`, and the manifest has the property `Multi-Release : true`. JDK8 and less will never load the JDK+ shim.

There is an example of how to do this for a JDK9+ vectorised `Arrays` equality method in [RoaringBitmap](https://github.com/RoaringBitmap/RoaringBitmap/tree/master/shims).

jctools-core/src/main/java/org/jctools/util/ThreadHints.java
```diff
@@ -0,0 +1,67 @@
...
+            {
+                okToInvokeMHCaller = false;
+            }
+        }
+        // okToInvokeThreadHintsMHCaller will be false for e.g. Java SE 6 and earlier:
+        okToInvokeInternalThreadHintsCaller = okToInvokeMHCaller;
+    }
+
+    // prevent construction...
+    private ThreadHints()
```
nitsanw: I used to do this too, but my current thinking is that it is pointless. The class is final, it has no fields, no methods that are not static. If some idiot wants to instantiate it, so what?

nitsanw:
In its current state this will not fly with building on 8 with JDK6 compatibility.

franz1981:
> In its current state this will not fly with building on 8 with JDK6 compatibility.

I know, I've yet to work on this to make this possible, but still not sure what to do: the idea of @richardstartin seems reasonable to me, but probably will need to support Multi-Release jar.

franz1981:
@nitsanw Nitsan, bud, I cannot add any label to this, can you do it for me? is "DON'T MERGE IT YET" in red; I will change it accordly when I've a bit of time and will proceed on this :+1:

richardstartin:
There's a demo of how to do this with Maven [here](https://github.com/richardstartin/multi-release).

kermandev:
Is this going to be revisted with the jdk11 module?

nitsanw:
This can be re-evaluated now that the base JDK requirement is 8, and we have a jdk11 module. @franz1981 @kermandev @richardstartin

Issue #289, opened 2020-03-08
Add Automatic-Module-Name header to MANIFEST.MF


rgoers:
Log4j 2 uses jctools-core as an optional dependency. On of the goals of Log4j 2 3.x is to fully support the Java Platform Module System. In order to do this it needs to declare all the modules (i.e. dependencies) it references. The best way to comply with this is to add a module-info.class file to the delivered jar. But if that is too difficult then adding the Automatic-Module-Name header to the META-INF/MANIFEST.MF needs to be done. Without this JPMS will use the name of the jar, minus its version number, as the module name, which will frequently cause problems and almost never will match the name you would want to use when fully modularized.

nitsanw:
PR is welcome :-)

rgoers:
See https://github.com/JCTools/JCTools/pull/290. It should fix this.

nitsanw:
Thanks!

PR #290, opened 2020-03-12
#289 Add Automatic-Module-Name header to MANIFEST.MF

vy:


vy:
@rgoers, would you mind checking if this is what you were looking for, please?

rgoers:
Yes, this should correct the problem.

nitsanw:
Thanks!

PR #295, opened 2020-05-22
Fix peek/relaxedPeek implementation of MpmcArrayQueue.

hl845740757:
Summary
**Poll/relaxedPoll/Peek/RelaxedPeek should not return an older value.**

ConsumerC: prepare to load element (currentConsumerIndex)  
ConsumerA: poll the element (currentConsumerIndex)  
Producer: offer the element (currentConsumerIndex + capaity)  
Producer: complete store element  
ConsumerC: complete load element (currentConsumerIndex + capaity)  

ConsumerC expects to load currentConsumerIndex, but loads (currentConsumerIndex + capaity).

The implementation of **SpmcArrayQueue** has the same bug. 

```
Exception in thread "Thread-2" java.lang.IllegalStateException: peekedSequence 18, lastPeekedSequence 25
	at com.wjybxx.fastjgame.concurrenttest.SpmcArrayQueuePeekTest$Peeker.run(SpmcArrayQueuePeekTest.java:111)
```

```
public class SpmcArrayQueuePeekTest {

    private static volatile boolean stop = false;

    public static void main(String[] args) throws InterruptedException {
        // Smaller capacity helps test
        MessagePassingQueue<Long> messageQueue = new SpmcArrayQueue<>(8);
        new Producer(messageQueue).start();
        new Consumer(messageQueue).start();
        new Peeker(messageQueue).start();

        try {
            Thread.sleep(10 * 1000);
        } finally {
            stop = true;
        }
    }

    private static class Producer extends Thread {

        final MessagePassingQueue<Long> messageQueue;

        long sequence = 0;

        Producer(MessagePassingQueue<Long> messageQueue) {
            this.messageQueue = messageQueue;
        }

        @Override
        public void run() {
            while (!stop) {
                if (messageQueue.offer(sequence)) {
                    sequence++;
                }
            }
        }
    }

    private static class Consumer extends Thread {

        final MessagePassingQueue<Long> messageQueue;

        private Consumer(MessagePassingQueue<Long> messageQueue) {
            this.messageQueue = messageQueue;
        }

        @Override
        public void run() {
            while (!stop) {
                messageQueue.poll();
                // wait producer fill
                try {
                    Thread.sleep(1);
                } catch (InterruptedException e) {
                    e.printStackTrace();
                }
            }
        }

    }

    private static class Peeker extends Thread {

        final MessagePassingQueue<Long> messageQueue;
        long lastPeekedSequence;

        private Peeker(MessagePassingQueue<Long> messageQueue) {
            this.messageQueue = messageQueue;
        }

        @Override
        public void run() {
            while (!stop) {
                final Long peekedSequence = messageQueue.peek();
                if (peekedSequence == null) {
                    continue;
                }

                if (peekedSequence < lastPeekedSequence) {
                    String msg = String.format("peekedSequence %s, lastPeekedSequence %s", peekedSequence, lastPeekedSequence);
                    throw new IllegalStateException(msg);
                }

                lastPeekedSequence = peekedSequence;
            }
        }
    }
}
```


nitsanw:
Thanks for the interesting test case. Can you add it to the existing test suite so all the queues are covered?

jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java
```diff
@@ -273,16 +273,30 @@ public E poll()
...
         do
         {
+            // Order matters!
             cIndex = lvConsumerIndex();
-            // other consumers may have grabbed the element, or queue might be empty
-            e = lpRefElement(buffer, calcCircularRefElementOffset(cIndex, mask));
-            // only return null if queue is empty
+            // Volatile Mode: Loading element must be completed before loading pIndex.
+            e = lvRefElement(buffer, calcCircularRefElementOffset(cIndex, mask));
+            pIndex = lvProducerIndex();
```
nitsanw, line 284: The original solution relying on the matching sequence is more in line with how `poll` works and the spirit of this queue altogether. The idea here is to avoid as much as possible cache misses/traffic induced by loading the producer index from the consumer threads.

jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java
```diff
@@ -350,8 +364,24 @@ public E relaxedPoll()
     @Override
     public E relaxedPeek()
     {
-        long currConsumerIndex = lvConsumerIndex();
-        return lpRefElement(buffer, calcCircularRefElementOffset(currConsumerIndex, mask));
+        // Order matters!
+        long cIndex = lvConsumerIndex();
+        // Volatile Mode: Loading element must be completed before loading pIndex.
+        E e = lvRefElement(buffer, calcCircularRefElementOffset(cIndex, mask));
+        long pIndex = lvProducerIndex();
```
nitsanw, line 371: relaxed peek should be implemented without loading the producer index at all.

jctools-core/src/main/java/org/jctools/queues/MpmcArrayQueue.java
franz1981: The sequence could be read twice to check if the read element is stable, similarly to what a stamped lock does

hl845740757: Stable value is unnecessary

nitsanw:
I think this is now resolved

PR #299, opened 2020-05-23
remove some redundant store fence

hl845740757:
For MpmcArrayQueue,  the producer rely on sequcne.  
For MpscArrayQueue, the prodecer rely on consumerIndex.


nitsanw:
Thanks!

nitsanw:
@hl845740757 please note that this commit broke the build, which I should have picked up, but highlights that you are not running the build locally before committing/pushing

PR #311, opened 2020-06-03
Fixes #310: MpmcUnboundedXaddArrayQueue#peek/relaxedPeek can load "future" elements

### LINKED ISSUE #310: MpmcUnboundedXaddArrayQueue#peek/relaxedPeek  can load "future" elements
If the speed of the producer is too fast, the producer always tends to create new chunks, so we must reduce the speed of the producer.
If the speed of the consumer is too fast, the peeker always tends to read an newer element, so we must reduce the speed of the consumer.

If and only if all chunks form a ring buffer, peeker may read an future element.
@franz1981 

```
Exception in thread "Thread-2" java.lang.IllegalStateException: peekedSequence 9201, lastPeekedSequence 9202
	at com.wjybxx.fastjgame.queue.MpmcUnboundedXaddArrayQueuePeekTest$Peeker.run(MpmcUnboundedXaddArrayQueuePeekTest.java:121)
```

Test case
```
public class MpmcUnboundedXaddArrayQueuePeekTest {

    private static final int chunkSize = 16;
    private static final int poolSize = 4;
    private static final int capacity = chunkSize * poolSize;

    private static volatile boolean stop = false;

    public static void main(String[] args) throws InterruptedException {
        MessagePassingQueue<Long> messageQueue = new MpmcUnboundedXaddArrayQueue<>(chunkSize, poolSize);
        new Producer(messageQueue, 800).start();
        new Consumer(messageQueue, 600).start();
        new Peeker(messageQueue).start();

        try {
            Thread.sleep(10 * 1000);
        } finally {
            stop = true;
        }
    }

    private static class Producer extends Thread {

        final MessagePassingQueue<Long> messageQueue;
        final long sleepNanos;

        long sequence = 0;

        Producer(MessagePassingQueue<Long> messageQueue, long sleepNanos) {
            this.messageQueue = messageQueue;
            this.sleepNanos = sleepNanos;
        }

        @Override
        public void run() {
            while (!stop) {
                if (messageQueue.offer(sequence)) {
                    sequence++;
                }

                if (messageQueue.size() >= capacity - chunkSize) {
                    LockSupport.parkNanos(sleepNanos);
                }
            }
        }
    }

    private static class Consumer extends Thread {

        final MessagePassingQueue<Long> messageQueue;
        final long sleepNanos;

        private Consumer(MessagePassingQueue<Long> messageQueue, long sleepNanos) {
            this.messageQueue = messageQueue;
            this.sleepNanos = sleepNanos;
        }

        @Override
        public void run() {
            while (!stop) {
                messageQueue.poll();

                if (messageQueue.size() < chunkSize) {
                    LockSupport.parkNanos(sleepNanos);
                }
            }
        }

    }

    private static class Peeker extends Thread {

        final MessagePassingQueue<Long> messageQueue;
        long lastPeekedSequence;

        private Peeker(MessagePassingQueue<Long> messageQueue) {
            this.messageQueue = messageQueue;
            setPriority(MIN_PRIORITY);
        }

        @Override
        public void run() {
            while (!stop) {
                final Long peekedSequence = messageQueue.peek();
                if (peekedSequence == null) {
                    continue;
                }

                if (peekedSequence < lastPeekedSequence) {
                    String msg = String.format("peekedSequence %s, lastPeekedSequence %s", peekedSequence, lastPeekedSequence);
                    throw new IllegalStateException(msg);
                }

                lastPeekedSequence = peekedSequence;
            }
        }
    }
}
```

franz1981:


franz1981:
@hl845740757 I see that this isn't easy to be reproduced so thanks for the effort of searching a viable test: I'm not happy with this PR due to the hardcoded time to test (ie 2 seconds in my case), but it should be reproducible in many boxes (I've tried with 3 different ones, including a large Xeon server).
The solution I've built is not perfect, because for maxPooledChunks == 0 and when the chunks are not pooled (both the one used to read next and the element itself) there is no need to read the consumer index againg.
Re this solution itself, probably I could have used the chunk Index or the sequence Index instead of the consumer index to check the stability of the element read too (this was my original implementation of peek > 1 year ago).

@nitsanw Please take a look bud and I will try to improve where we need it ;)

jctools-core/src/test/java/org/jctools/queues/MpqSanityTestMpmcUnboundedXadd.java
```diff
@@ -21,13 +26,178 @@ public MpqSanityTestMpmcUnboundedXadd(ConcurrentQueueSpec spec, MessagePassingQu
...
         list.add(makeMpq(0, 0, 0, Ordering.FIFO, new MpmcUnboundedXaddArrayQueue<>(1, 3)));
-        list.add(makeMpq(0, 0, 0, Ordering.FIFO, new MpmcUnboundedXaddArrayQueue<>(64, 3)));
+        list.add(makeMpq(0, 0, 0, Ordering.FIFO, new MpmcUnboundedXaddArrayQueue<>(16, 3)));
+        list.add(makeMpq(0, 0, 0, Ordering.FIFO, new MpmcUnboundedXaddArrayQueue<>(1, 4)));
+        list.add(makeMpq(0, 0, 0, Ordering.FIFO, new MpmcUnboundedXaddArrayQueue<>(16, 4)));
         return list;
     }
+
+    @Test
+    public void peekShouldNotSeeFutureElements() throws InterruptedException
```
nitsanw, line 42: Can you use the threads infra in existing tests?

nitsanw:
Thanks guys

PR #312, opened 2020-06-18
Add a wait-free intrusive stack data structure

chrisvest:
The WFIStack is added to the set of experimental data structures.
This data structure has, from experience, proven useful for implementing a number of other concurrency primitives.

Let me know if this is interesting.

### DIFF
--- a/jctools-experimental/src/main/java/org/jctools/stacks/WFIStack.java
+++ b/jctools-experimental/src/main/java/org/jctools/stacks/WFIStack.java
@@ -13,6 +13,7 @@
  */
 package org.jctools.stacks;
 
+import org.jctools.stacks.WFIStack.Node;
 import org.jctools.util.UnsafeAccess;
 
 import java.util.Iterator;
@@ -36,7 +37,7 @@ import java.util.NoSuchElementException;
  * @param <T> The type of node in this stack.
  */
 @SuppressWarnings({"unchecked", "NullableProblems"})
-public class WFIStack<T extends WFIStack.Node> implements Iterable<T>
+public class WFIStack<T extends WFIStack.Node> extends PadAfter implements Iterable<T>
 {
     /**
      * The super-class of all nodes, or entries, for {@link WFIStack wait-free intrusive stacks}.
@@ -49,10 +50,10 @@ public class WFIStack<T extends WFIStack.Node> implements Iterable<T>
         final Node next()
         {
             Node n = next;
+            //noinspection IdempotentLoopBody
             while (n == null)
             {
                 // TODO: Thread.onSpinWait
-                Thread.yield();
                 n = next;
             }
             return n;
@@ -135,10 +136,6 @@ public class WFIStack<T extends WFIStack.Node> implements Iterable<T>
             return null;
         }
     };
-    private static final long HEAD_OFFSET = UnsafeAccess.fieldOffset(WFIStack.class, "head");
-
-    @SuppressWarnings( {"FieldCanBeLocal", "FieldMayBeFinal"})
-    private volatile Node head; // Accessed via UNSAFE.
 
     /**
      * Create an empty wait-free intrusive stack.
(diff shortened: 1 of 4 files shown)


jctools-experimental/src/main/java/org/jctools/stacks/WFIStack.java
```diff
@@ -0,0 +1,329 @@
...
+        volatile Node next;
+        int count;
+
+        final Node next()
+        {
+            Node n = next;
+            while (n == null)
+            {
+                // TODO: Thread.onSpinWait
+                Thread.yield();
```
franz1981, line 55: Looking at the Vyukov MpscLinkedQueue poll we tends to not use Thread::yield although I understand that with a single threaded machine could be the right way to handle the "bubble" while awaiting next to materialize: I would left a raw spin wait here

jctools-experimental/src/main/java/org/jctools/stacks/WFIStack.java
```diff
@@ -0,0 +1,329 @@
...
+        @Override
+        <X extends Node> X self()
+        {
+            return null;
+        }
+    };
+    private static final long HEAD_OFFSET = UnsafeAccess.fieldOffset(WFIStack.class, "head");
+
+    @SuppressWarnings( {"FieldCanBeLocal", "FieldMayBeFinal"})
+    private volatile Node head; // Accessed via UNSAFE.
```
franz1981, line 141: I'm not a big fan of padding, especially manual padding, but probably I would pad this from any heap noisy neighbours, including the obj header itself

jctools-experimental/src/main/java/org/jctools/stacks/WFIStack.java
```diff
@@ -0,0 +1,329 @@
...
+ */
+package org.jctools.stacks;
+
+import org.jctools.util.UnsafeAccess;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+/**
+ * A Wait-Free Intrusive Stack (Last-In, First-Out).
```
franz1981, line 22: Do you think would be nice to implement an additional `default` non-intrusive version (a-la CLQ) based on this for lazy users?

jctools-experimental/src/test/java/org/jctools/stacks/WFIStackTest.java
```diff
@@ -0,0 +1,395 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+import static org.junit.Assert.*;
+
+public class WFIStackTest
```
franz1981, line 24: We've been recently bitten by some unexpected concurrent behaviour of the JCTools data structures so I suggest to attempt writing some mpmc variation of tests although I understand the pain of making them reproducible/usefull

chrisvest: What's a good example to follow for writing such a test?

jctools-experimental/src/main/java/org/jctools/stacks/WFIStack.java
chrisvest: What's the CLQ you're referring to?

jctools-experimental/src/test/java/org/jctools/stacks/WFIStackTest.java
franz1981: This one is a possible way to write it: https://github.com/JCTools/JCTools/blob/c0b570d31a4a60cb2b538f86106207f626c4b30c/jctools-core/src/test/java/org/jctools/queues/QueueSanityTest.java#L556

Using sequenced events to track expected order

jctools-experimental/src/main/java/org/jctools/stacks/WFIStack.java
franz1981: Sorry, you're right, I've missed to translate the acronym: `ConcurrentLinkedQueue` of the j.u.c

nitsanw:
Thanks @chrisvest ! this is interesting, I'll have a close read next week :-)

jctools-experimental/src/main/java/org/jctools/stacks/WFIStack.java
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    private boolean casHead(Node expected, Node update)
+    {
+        return UnsafeAccess.UNSAFE.compareAndSwapObject(this, HEAD_OFFSET, expected, update);
+    }
+}
+
+@SuppressWarnings("unused")
+abstract class PadBefore
```
nitsanw, line 339: Please reorder and rename padding classes as per other classes in JCTools. i.e. padding in order of inheritance, padding fiends copy and pasted from existing code(helps searching)

jctools-experimental/src/main/java/org/jctools/stacks/WFIStack.java
```diff
@@ -0,0 +1,373 @@
...
+    {
+        assert node.next == null : "WFIStack.Nodes cannot be reused.";
+    }
+
+    private T xchgHead(Node node)
+    {
+        return (T) UnsafeAccess.UNSAFE.getAndSetObject(this, HEAD_OFFSET, node);
+    }
+
+    private boolean casHead(Node expected, Node update)
```
nitsanw, line 332: Convention in the project is to put the accessors next to the field.

jctools-experimental/src/main/java/org/jctools/stacks/WFIStack.java
```diff
@@ -0,0 +1,373 @@
...
+        checkPush(node);
+        node.count = 1;
+        node.next = END;
+        return new NodeIterable<T>(xchgHead(node)).reverse();
+    }
+
+    /**
+     * Pop, or remove, the most recently pushed node, or return {@code null} if the stack is empty.
+     * <p>
+     * This operation is non-blocking, but not wait-free. It has constant time-complexity.
```
nitsanw, line 297: I would say it's lock-free (in the "WF > LF > blocking" sense). I'm also not sure the "constant time complexity" is informative or correct here. The worst case time here is infinite through no relation to N IIUC.

chrisvest:
Currently working on adding a benchmark.

PR #313, opened 2020-06-27
Flat Combining universal construct

chrisvest:
This PR is using #312 as a base line. Work harder, @nitsanw 😛 

This is not yet ready to merge, I think. But it is ready for discussing the API. I'm on the fence about using type parameters for exceptions. Perhaps there should be more methods and operation types with different signatures that would make Java 8 method references more easily align with common signature patterns seen in the Java collections framework. It's also possible, I think, to separate the enqueueing from the execution, and have an apply method that returns a Future.

### DIFF
--- a/jctools-experimental/src/main/java/org/jctools/uc/FlatCombining.java
+++ b/jctools-experimental/src/main/java/org/jctools/uc/FlatCombining.java
@@ -1,0 +1,181 @@
+package org.jctools.uc;
+
+import org.jctools.stacks.WFIStack;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.LockSupport;
+
+/**
+ * FlatCombining is a <em>universal construct</em> which can provide thread-safe and concurrent access to an
+ * <em>arbitrary</em> sequential data structure.
+ * <p>
+ * The only requirement is that all accesses to the data structure must be delivered as
+ * {@linkplain Operation operations} through the {@link #apply(Operation)} method.
+ * <p>
+ * Each operation will only execute exactly once, and there will only be a single copy of the underlying data structure.
+ * <p>
+ * The only consequcne is that the {@link #apply(Operation)} method may block.
+ * <p>
+ * This is conceptually similar to wrapping the data structure in a mutex, but the {@code FlatCombining} implementation
+ * will scale positively with the level of concurrency, while contended locks typically scales negatively.
+ * Negative scalability is when the system-wide throughput drops as the number of active threads increases.
+ * <p>
+ * The Flat Combining concept is described by the paper
+ * <a href="https://www.cs.bgu.ac.il/~hendlerd/papers/flat-combining.pdf">
+ * Flat Combining and the Synchronization-Parallelism Tradeoff</a>
+ *
+ * @param <T> The type of data managed by this flat combining instance.
+ */
+public class FlatCombining<T> {
+    private final WFIStack<OpNode> ops;
+    private final AtomicBoolean lock;
+    private final T data;
+
+    public FlatCombining(T sequentialData) {
+        ops = new WFIStack<OpNode>();
+        lock = new AtomicBoolean();
+        data = sequentialData;
+    }
+
+    /**
+     * Apply the given operation on the underlying data structure. This call may block. Multiple threads can calls
+     * this at the same time. The operations will be applied with linearizable consistency.
+     * <p>
+     * The operation will only be applied once to the data structure, but there is no guarantee that it will be
+     * applied by the calling thread.
+     *
+     * @param operation The operation to apply to the underlying data structure.
+     * @param <R>       The type of return value produced by the operation, if any.
+     * @param <E>       The type of exception the operation might throw, if any.
+     * @return The result returned by the operation.
+     * @throws E                    If the operation throws an exception.
+     * @throws InterruptedException If the apply method ends up blocking but the thread is interrupted.
+     */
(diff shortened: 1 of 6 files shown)


jctools-experimental/src/main/java/org/jctools/uc/FlatCombining.java
```diff
@@ -0,0 +1,181 @@
...
+                }
+            }
+        }
+    }
+
+    private <R, E extends Exception> OpNode enqueue(Operation<T, R, E> operation) {
+        if (operation == null) {
+            throw new NullPointerException("The operation cannot be null.");
+        }
+        OpNode op = new OpNode(operation);
```
franz1981, line 83: 2 comments:
# I often see combiners using a thread local reusable node to save allocating in the hot path, do you think could worth doing it? Given that execution completed mark the end of the node lifecycle it should make possible to pool it
# is see we poll with FIFO policy, why not leaving the choice of using different data-structures (eg lifo/fifo) to store the ops?

jctools-experimental/src/main/java/org/jctools/uc/FlatCombining.java
```diff
@@ -0,0 +1,181 @@
...
+        ops.push(op);
+        return op;
+    }
+
+    private boolean tryLock() {
+        return !lock.get() && lock.compareAndSet(false, true);
+    }
+
+    private Iterable<OpNode> combineAndApplyAllOperations() {
+        Iterable<OpNode> nodes = ops.popAllFifo();
```
franz1981, line 93: This is a special single consumer operation, because protected by the lock ie the stack could expose a lighter single consumer (or exclusive consumer) variant of this (or a weakPopAllFifo that can fail under contention too; a condition they we won't expect to happen here ;))

jctools-experimental/src/main/java/org/jctools/uc/FlatCombining.java
chrisvest: We could do pooling. An apply method that separates enqueueing and application in time via a Future can't rely on pooling, though. Those operations will have to be unpooled. Perhaps they need to be in a separate stack as well to prevent the pooled stack from growing too big. The lock could be turned from a boolean to an int, and include a sequence so we can periodically truncate the stack and avoid leaking.

Applying the operations in FIFO order is how the linearisable consistency is achieved. Would be easy to add an option to turn that off.

jctools-experimental/src/main/java/org/jctools/uc/FlatCombining.java
chrisvest: You're imagining a `tryApply` method that only applies if it can do so without blocking?

I'm not a fan of that, since we have to enqueue before we can make an attempt at the lock. If getting the lock fails, we'd have cancellation racing with another thread applying our operation. Even if that gets coordinated, we'd still leave our operation enqueued until another thread comes to apply.

chrisvest: For instance, we don't know how long an operation is going to take, and once it has been _started_ we can't return from `apply` until it's been completed. Otherwise I think we'll get a programming model that's hard to reason about.

franz1981: You're right :)
I've now looked how the stack is implementing `popAllFifo` and I've forgotten that's not a queue ie it doesn't have separated head and tail pointers, hence cannot implement a `single-writer` variant of poll.
Instad, I see that `MpscLinkedQueue` already implement a single-consumer FIFO (by definition) poll and has a wait-free `enqueue` as well: maybe the `FlatCombining` construct could abstract which `fifo` data-structure to use eg `MpscLinkedQueue`,`WFIStack` or other...

jctools-experimental/src/main/java/org/jctools/uc/FlatCombining.java
chrisvest: Thinking about the consistency... since concurrent blocking apply calls overlap in time, I think we can pick an arbitrary ordering of these overlapping operations, and still be considered linearisable. The operations were already racing, so there's no externally observable sequence of events that can disagree with applying the operations in LIFO order.

jctools-experimental/src/main/java/org/jctools/uc/FlatCombining.java
```diff
@@ -0,0 +1,181 @@
...
+     * @return The result returned by the operation.
+     * @throws E                    If the operation throws an exception.
+     * @throws InterruptedException If the apply method ends up blocking but the thread is interrupted.
+     */
+    public <R, E extends Exception> R apply(Operation<T, R, E> operation) throws E, InterruptedException {
+        OpNode op = enqueue(operation);
+
+        for (; ; ) {
+            if (tryLock()) {
+                Iterable<OpNode> nodes = combineAndApplyAllOperations();
```
franz1981, line 59: Although counter-intuitive I think that would be nice to have a version of `combineAndApplyAllOperations` that allow to limit the amount of work to be performed (the simpler way in the number of combined operations, but can think to other metrics, or by using a cooperative suspention of some form eg `ControlledRunnable`?) to allow other(s) Threads to pick remaining jobs (if the combiner thread isn't alone).

franz1981: Thinking it twice probably the limit in the combined list of pending ops work only if the list is FIFO. Need to think about it better

chrisvest:
How about we close this and #312? I don't think there's any demand for these, so no point in keeping them around.

PR #314, opened 2020-07-10
MpscBlockingConsumerArrayQueue.offerIfBelowThreshold()

philipa:
Add offerIfBelowThreshold() to MpscBlockingConsumerArrayQueue, with the same semantics as MpscArrayQueue.offerIfBelowThreshold().

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
@@ -297,7 +297,7 @@ public class MpscBlockingConsumerArrayQueue<E> extends MpscBlockingConsumerArray
      * {@link #offer} if {@link #size()} is less than threshold.
      *
      * @param e         the object to offer onto the queue, not null
-     * @param threshold the maximum allowable size; [1 <= threshold <= capacity].
+     * @param threshold the maximum allowable size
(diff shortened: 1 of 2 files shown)


nitsanw:
Testing needs to be extended, see the method used for the same method on MpscArrayQueue. We also need to cover the scenarios around unparking the consumer here, which will require some refactoring to the extended cases I expect.

nitsanw:
See suggested changes here: https://github.com/JCTools/JCTools/commit/9e2d88242f009de02edc32c919622a2f5d313876

jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
```diff
@@ -293,6 +293,76 @@ public String toString()
...
+            }
+            // pIndex is even (lower bit is 0) -> actual index is (pIndex >> 1), consumer is awake
+            final long producerLimit = lvProducerLimit();
+
+            // Use producer limit to save a read of the more rapidly mutated consumer index.
+            // Assumption: queue is usually empty or near empty
+
+            final long available = (producerLimit - pIndex) >> 1;
+            // sizeEstimate <= size
+            final long sizeEstimate = capacity - available;
```
nitsanw, line 334: This is a bug as `capacity` is actually `2xcapacity`

PR #316, opened 2020-07-20
MpscBlockingConsumerArrayQueue poll with timeout fix

philipa:
Reorder offerAndWakeup() to fix race. Previously the following sequence was possible:

    Consumer                            Provider

    poll(timeout)
                                      offerAndWakeup()
    time out
                                      soRefElement(, offset,)
    consume from offset
    offset -> offset'
    poll(timeout)
                                      soBlocked(null)
                                      unpark(consumerThread)
    // Spin for ever
    spinWaitForElement(, offset')


franz1981:
I don't understand why the last poll should spin wait for element TBH,  given that there is no new element on  offset' beside the one already consumed by take(timeout), need to look better into this

philipa:
Look at `spinWaitForElement`. It will wait for ever for an update to `offset`, but none is forthcoming. Further, the "busy bit" of `producerIndex` is left set but `blocked` is `null`, so the next `offer` will spin too.

Or try reverting the fix and running the test. It failed consistently for me, even with additional logging so I could figure out what was happening.

franz1981:
What it looks strange to me is the check on https://github.com/pushtechnology/JCTools/blob/865a75f12cfa3d63029f16979dc8ee2af7e4a5b1/jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java#L577 : I will add some logs to understand why it isn't returning null instead

philipa:
Oh, right. I was using `poll(timeout)` throughout, so this is irrelevant. I'll updated the comment above accordingly.

franz1981:
Ah ok so the issue happens with 2 subsequent poll(timeout) and the expectation would be that:
1) first one should time out but read the offered element 
2) second one should time out and return null

Or

1) first one should time out and return null
2) second one should return an element regardless timing out (if the offer land in time)

philipa:
Yes, at least two `poll(timeout)`s and a concurrent `offerAndWait()`.

It reproduced for me fairly readily once I had the test balanced to ensure a mostly empty queue. I'd be interested to know if the test fails for you (with the fix reverted).

nitsanw:
Hi @philipa ! long time no see :-)

Thanks for the find + PR, this is an interesting one. The thinking behind the original ordering is that we want to avoid the spin when we wake the consumer from an unbounded `take`. There have been prior concerns around spinning with no `yield` in single CPU environments or where the scheduler `unpark` results in the unparked thread executing on the same core. Upon reflection I'm not sure if re-ordering is the "right" solution (I think it's correct), or if we need heavier weight guards on the state transitions here via e.g. a CAS on the blocked field. WDYT?

philipa:
Not sure CAS on blocked works alone, since the producer would need to distinguish between the consumer waiting for a value at the original offset vs it having woken up early, sneakily read the element, and now waiting for an element at the next offset. `cIndex` is also pertinent here.

njhill:
Here is an alternative fix to consumer-side which should not affect the existing happy paths at all: 3f82f1c19f9b2886f4a5738309d40bd3e8a5412f. The existing cas in the finally block already detects this race and so it's just a matter of handling it appropriately (I removed the try/catch and now have separate checks for interrupt and timeout cases). wdyt?

philipa:
Looks good. I prefer the revised code flow. It would make sense to restructure `take` similarly to remove the finally block.

I did like e8e0839b16f2ef5ee37676910d7ab22360e66798 though.

franz1981:
+100 for the changes of @njhill

njhill:
Thanks @philipa @franz1981, agree about the restructure of `take`. I should note that I was the one that introduced this bug in the first place, as well as the the try/finally flow which @nitsanw [wasn't in favour of](https://github.com/JCTools/JCTools/pull/247#discussion_r293311707) at the time... so more reason to change it I guess!

jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
```diff
@@ -428,32 +424,22 @@ public E take() throws InterruptedException
...
-                }
-                finally
-                {
-                    if (!unblocked) {
                         // revert blocking state
                         if (casProducerIndex(pIndex + 1, pIndex))
                         {
                             soBlocked(null);
                         }
+                        throw new InterruptedException();
```
njhill, line 439: @philipa I think the cas-fail case should be handled differently here - to basically just ignore the interruption but don't reset the interrupt status then continue spinning for unblock and then return the element instead of throwing, like I did in the `take(timeout)` case below. Otherwise this element will be "lost"

philipa: How is the element lost? We've not advanced `cIndex` and the producer is in control of `pIndex`.

njhill: @philipa oops yes you're right, my bad... so then it would make sense for the interrupt/timeout logic in `take(timeout)` to also match this for simplicity.

However I think you do still need to spin-wait to see `unblocked` null prior to returning/throwing in all these cas failure cases or else there is still a race possible.

philipa: I don't think the spin-wait is required. The producer has the responsibility of unblocking and cleaning up. Isn't identical handling sufficient?

                    if (Thread.interrupted())
                    {
                        if (casProducerIndex(pIndex + 1, pIndex)) {
                            soBlocked(null);
                        }
                        throw new InterruptedException();
                    }

njhill: @philipa what about:

C: interrupted in `take()`
P1: offers, wins producer index race, `soRefElement(A)`
C: throws IE
C: `take()` again consumes `A`
C: `get(timeout)`, cas `soBlocked(currentThread)`, enter park loop
P1: soBlocked(null), unpark
C: breaks from loop, enters `spinWaitForElement`
timeout elapses or C is interrupted, C does not return (at least until further producer activity)

philipa: I had to stare at this, but .... agreed. It's similar to the original race. It could also happen using `take()` throughout.

So the consumer should always `spinWaitForUnblocked()` if it fails to revert the blocking state. I'll do that.

nitsanw:
@philipa thanks for the good work! I'll do a final pass today

nitsanw:
@philipa @njhill See minor changes: https://github.com/pushtechnology/JCTools/commit/c3ea43356d7a50e47b92f97c5b61463653252c13

If you can squash it all to one nice commit I think we're done :-)

philipa:
Refactoring looks good.

Excuse my GH ignorance, but don't you have a "squash and merge" button for PRs?

nitsanw:
Great work team :-)

PR #323, opened 2020-10-14
Update to JCStress 0.7 and fix tests for updated APIs

shipilev:
JCTools use the ancient JCStress 0.2. Let's use a much more modern 0.7.

Testing:
  - [x] `mvn clean install -pl jctools-concurrency-test`
  - [x] `java -jar jctools-concurrency-test/target/concurrency-test.jar` (`SingleWriterHashSetDuplicateReadsTest` still fails, as it should?).

### DIFF
--- a/jctools-concurrency-test/pom.xml
+++ b/jctools-concurrency-test/pom.xml
@@ -12,7 +12,7 @@
     <name>concurrency-test</name>
 
     <properties>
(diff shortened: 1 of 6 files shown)


franz1981 (approved):
LGTM

nitsanw:
Thanks for the contribution @shipilev !

PR #326, opened 2020-10-29
Xadd queues consumers can help producers

franz1981:


franz1981:
@nitsanw I've yet to decide if relaxedXXX makes sense will do this too (especially for the mpsc variant)

franz1981:
This one should be code completed: probably is time for me to fix the generator to creat the atomic variant

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
```diff
@@ -216,6 +216,20 @@ private E linkNextConsumerChunkAndPoll(
         MpmcUnboundedXaddChunk<E> next,
         long expectedChunkIndex)
     {
+        if (next == null) {
+            final long cChunkIndex = expectedChunkIndex - 1;
+            assert cChunk.lvIndex() == cChunkIndex;
+            // no need to try more then once here:
+            // if there is already a rotation in place or
```
nitsanw, line 223: I'm not sure what is guiding your line length here... Is this a poem? :P

franz1981: I admit it's a bit verbose :P
And I've read more inspired ones too, so I can rephrase it with  TLDR version :D

franz1981: mmm but you agree on the logic? There are other queues that can/should get same treatment? (Not the poem :P)

nitsanw:
reading through it now

nitsanw (approved):
LGTM, please consider suggestions

jctools-core/src/main/java/org/jctools/queues/MpmcUnboundedXaddArrayQueue.java
```diff
@@ -216,6 +216,14 @@ private E linkNextConsumerChunkAndPoll(
         MpmcUnboundedXaddChunk<E> next,
         long expectedChunkIndex)
     {
+        if (next == null) {
+            final long cChunkIndex = expectedChunkIndex - 1;
```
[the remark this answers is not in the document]
franz1981: Taken both the good advices, thanks :)

I've changed back the mpsc method name instead because the fact that `spinForNextIfNotEmpty` is trying append a new chunk is just an implementation detail and I'm not sure it should leak in the method name

Issue #319, opened 2020-08-13
Non-linearizable return value of put operation in NonBlockingHashMapLong


alefedor:
There is a bug in jctools 3.1 that results in `NonBlockingHashMapLong.put` returning not the previous value, but the next one added in a concurrent thread.

The bug can be reproduced with [Lincheck](https://github.com/Kotlin/kotlinx-lincheck).
The scenario with incorrect results has two parallel threads with one operation each, and one operation after the threads are finished:
```
Parallel part:
| put(5, -2): null | put(5, -8): -2 |
Post part:
[get(5): -2]
```
If `put(5, -8)` returned `-2`, then the result of the next `get` should have been `-8`.

There was the same bug in jctools 2.1 for `NonBlockingHashMap`, which seem to be fixed. 

In case the reasons of the bug differ from `NonBlockingHashMap`, I also attach the trace with only two context switches that leads to non-linearizable results (yet to be released feature in Lincheck).
```
= Parallel part execution: =
|                      | put(5, -8)                                                                                            |
|                      |   _chm.READ: @313a67ba at NonBlockingHashMapLong.putIfMatch(NonBlockingHashMapLong.java:316)          |
|                      |   READ: null at NonBlockingHashMapLong$CHM.putIfMatch(NonBlockingHashMapLong.java:559)                |
|                      |   switch                                                                                              |
| put(5, -2): null     |                                                                                                       |
|   thread is finished |                                                                                                       |
|                      |   READ: null at NonBlockingHashMapLong$CHM.putIfMatch(NonBlockingHashMapLong.java:560)                |
|                      |   CAS_val(5,null,-8): false at NonBlockingHashMapLong$CHM.putIfMatch(NonBlockingHashMapLong.java:641) |
|                      |   READ: null at NonBlockingHashMapLong$CHM.putIfMatch(NonBlockingHashMapLong.java:651)                |
|                      |   result: -2                                                                                          |
|                      |   thread is finished                                                                                  |
```

nitsanw:
Hi! Thanks for reporting! This looks like a useful test/tool, could you PR a working setup so we can add it to the release/build process?

alefedor:
@nitsanw 
Hi! We are about to release a new testing mode for Lincheck that can much more than just stress testing.
I'll make a PR after we finish the release.

nitsanw:
@alefedor that sounds awesome! thanks!

alefedor:
@nitsanw 
[...]

PR #328, opened 2020-11-16
Linearizability tests for concurrent maps and sets

alefedor:
This pull request adds linearizability tests for concurrent maps and sets, based on [Lincheck](https://github.com/Kotlin/kotlinx-lincheck).
Lincheck tests are written in a declarative style, for which you need only to declare possible concurrent operations in the data structure - Lincheck will generate concurrent scenarios, run them, and verify their results automatically.

The tests help to reproduce the bug in #319 with a trace available for an incorrect execution.

They also expose another bug that `NonBlockingIdentityHashMap.replace` fails if there were no such key before.
For example, `new NonBlockingIdentityHashMap().replace(1, 0);` will throw an `AssertionError`, because of unexpected `null` value. I have a strong feeling that a `TOMBSTONE` should have been there instead.

I also found that a `size` method is not linearizable when a concurrent `put` operation is in action. But I think that it is a feature rather than a bug, because _efficient and linearizable_ `size` operation is often difficult to implement, although I haven't found any note in the documentation about the problem.

### DIFF
--- a/jctools-core/pom.xml
+++ b/jctools-core/pom.xml
@@ -33,7 +33,24 @@
 			<version>${guava-testlib.version}</version>
 			<scope>test</scope>
 		</dependency>
+
+		<dependency>
+			<groupId>org.jetbrains.kotlinx</groupId>
+			<artifactId>lincheck</artifactId>
+			<version>${lincheck.version}</version>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
+
+	<repositories>
+		<repository>
+			<!-- Source for org.jetbrains.kotlinx.lincheck -->
+			<id>bintray-kotlin-kotlinx</id>
+			<name>kotlin-kotlinx</name>
+			<url>https://kotlin.bintray.com/kotlinx</url>
+		</repository>
+	</repositories>
+
 	<build>
 		<plugins>
 			<plugin>
--- a/jctools-core/src/main/java/org/jctools/maps/NonBlockingHashMapLong.java
+++ b/jctools-core/src/main/java/org/jctools/maps/NonBlockingHashMapLong.java
@@ -22,6 +22,7 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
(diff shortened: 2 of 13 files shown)


nitsanw:
This looks AWESOME!!! thanks for the contribution, I will have a look at it in the near future and integrate.

nitsanw:
@alefedor Can you please open "They also expose another bug that NonBlockingIdentityHashMap.replace fails if there were no such key before.
For example, new NonBlockingIdentityHashMap().replace(1, 0); will throw an AssertionError, because of unexpected null value. I have a strong feeling that a TOMBSTONE should have been there instead." as a separate bug? I'm not sure exactly what you mean here.

alefedor:
@nitsanw Ok. Opened the issue #330

ben-manes:
I was curious and tried this test against ConcurrentHashMap, but it fails on JDK16. Any insights on why?

```
= Invalid execution results =
Init part:
[put(4, 4): null]
Parallel part:
| get(1):    4    | put(1, 4): null |
| remove(1): null | clear():   void |
| get(4):    4    |                 |
```

alefedor:
Hi @ben-manes !

Nice observation, that's definitely a bug in `java.util.ConcurrentHashMap`.
Sometimes a result of a finished`clear` is seen by one read in a parallel thread, but is not seen by the next read.

Deducing from the fact that Lincheck can find this bug in the stress testing mode, but not in the model checking one, I suppose that it is related to weak memory model effects (e.g., missing memory barriers).

alefedor:
Okay, it may also be a feature, not a bug
```
For aggregate operations
such as {@code putAll} and {@code clear}, concurrent retrievals may
reflect insertion or removal of only some entries.
 ```
https://github.com/openjdk/jdk/blob/2ee1f96c14b80b63a29445629b1f2e1caf88e075/src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java#L92

ben-manes:
Ahh, that helps explain it then because a backport of CHMv7 also fails in a similar fashion. I was curious about adding the checks to [Caffeine](https://github.com/ben-manes/caffeine) which as an `asMap()` view. Maybe you could advise me off thread about what a good test suite would look like?

alefedor:
@ben-manes 

> Maybe you could advise me off thread about what a good test suite would look like?

Yes, sure. Email: afedorov2602@gmail.com

PR #336, opened 2021-03-02
MpscBlockingConsumerArrayQueue: fix race writing to blocked field

philipa:
Fix for this race:

```
Consumer in parkUntilNext()         Producer in offerAndWakeup()

casProducerIndex(0, 1)
                                    casProducerIndex(1, 2)
                                    soBlocked(null)
soBlocked(thread)

... spin until timeout

casProducerIndex(1, 0) fails

spinWaitForUnblock() for ever
```

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
@@ -423,8 +423,7 @@ public class MpscBlockingConsumerArrayQueue<E> extends MpscBlockingConsumerArray
         final long offset = modifiedCalcCircularRefElementOffset(pIndex, mask);
         final Thread consumerThread = lvBlocked();
 
-        // We could see a null here through a race with the consumer not yet storing the reference, or through a race
-        // with another producer. Just retry.
+        // We could see a null here through a race with the consumer not yet storing the reference. Just retry.
         if (consumerThread == null)
         {
             return false;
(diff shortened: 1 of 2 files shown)


jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
```diff
@@ -542,32 +531,36 @@ private E parkUntilNext(E[] buffer, long cIndex, long offset, long timeoutNs) th
...
-                        break; // just in the nick of time
+                    // ignore deadline when it's forever
+                    timeoutNs = timeoutNs == Long.MAX_VALUE ? Long.MAX_VALUE : deadlineNs - System.nanoTime();
+                    if (timeoutNs <= 0)
+                    {
+                        if (casProducerIndex(pIndex + 1, pIndex)) {
+                            // ran out of time and the producer has not moved the index
+                            return null;
+                        }
+                        else {
```
franz1981, line 555: no need to else here

franz1981 (changes requested):

jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
philipa: Done.

franz1981 (approved):
LGTM I haven't yet run the reproducer TBH, but the fix logic looks good

nitsanw:
Thanks @philipa this is a bad bug and a nice fix! will review and get a release out ASAP

PR #338, opened 2021-03-03
Fixes #337 Provide blocking peeks for MpscBlockingConsumerArrayQueue

### LINKED ISSUE #337: Provide blocking peeks for MpscBlockingConsumerArrayQueue
It should allow to implement batching draining strategies using a mix of blocking peek + drain 

eg
```java
// ...
do {
  while (q.blockingPeek(timeOut, unit) != null) {
      // same while using poll too
      if (q.drain(batch::offer, limit) > 0) {
          processBatch(batch);
      }
  }
  // perform timed check on something
} while (true); // or whatever other way to end this madness :P
```
Doing the same with the current API means using `take` or `poll` ie adding the just read `e` to the batch before using `drain` again, that's less "nice" (just a matter of taste, I know) and more verbose

franz1981:


franz1981:
@philipa too ^^

@nitsanw I'm going to remove the draft status if the idea is welcome and I'm going to add some tests too

jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
```diff
@@ -521,7 +554,7 @@ public E poll(long timeout, TimeUnit unit) throws InterruptedException
         return e;
     }
 
-    private E parkUntilNext(E[] buffer, long cIndex, long offset, long timeoutNs) throws InterruptedException {
+    private E parkUntilNext(E[] buffer, long cIndex, long offset, long timeoutNs, boolean peek) throws InterruptedException {
```
philipa, line 557: I'm not a fan of adding further complexity to `parkUntilNext`.

I was previously considering splitting out a version without the timeout for `take`; and removing the ternary checks for timeoutNs = Long.MAX_VALUE. (The downside to this is that you might have an unnecessary call  to System.nanoTime() if the poll timeout is greater than Long.MAX_VALUE ns ~ 300 years, but the caller can always use `poll()`).

If we did this, then we could also separate out and inline the appropriate consumption behaviour in the tail without the `peek` boolean.

franz1981: Given that's a supposed to be a slowpath method I didn't care much about adding a single check there, but I am more concerned about the added API; if it's welcome and make sense and if it can be useful to users. Any feedback re this?

Thanks for the review 😊

nitsanw:
I'm reluctant to add this. There's no blocking `peek` in https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html and the utility/need is unclear.

jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
philipa: I agree with the consensus on blockingDrain.

PR #341, opened 2021-03-03
Fixes #340 Provide blocking drain for MpscBlockingConsumerArrayQueue

### LINKED ISSUE #340: Provide blocking drain for MpscBlockingConsumerArrayQueue
This supersede https://github.com/JCTools/JCTools/issues/337

franz1981:


franz1981:
@nitsanw There is any chance we can get this one into the next release?

franz1981:
@nitsanw I can explode all the method calls and save `if (lvRefElement(buffer, offset) == null)` to happen (because `drain` has already returned 0, but that's the semantic I was thinking

jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
```diff
@@ -791,6 +793,48 @@ public int fill(Supplier<E> s, int limit)
...
+     */
+    public int drain(Consumer<E> c, final int limit, long timeout, TimeUnit unit) throws InterruptedException {
+        if (limit == 0) {
+            return 0;
+        }
+        final int drained = drain(c, limit);
+        if (drained != 0) {
+            return drained;
+        }
+        final E[] buffer = consumerBuffer;
```
nitsanw, line 819: I think you can just use `poll(timeout,unit)` here to pass one element to consumer, and follow up with second drain attempt as you do.

jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
```diff
@@ -567,7 +567,9 @@ private E parkUntilNext(E[] buffer, long cIndex, long offset, long timeoutNs) th
         // producer index is visible before element, so if we wake up between the index moving and the element
         // store we could see a null.
         e = spinWaitForElement(buffer, offset);
-
+        if (!consume) {
```
nitsanw, line 570: we don't need this is we use `poll`

jctools-core/src/main/java/org/jctools/queues/MpscBlockingConsumerArrayQueue.java
nitsanw:
```
E e = poll(timeout, unit);
if (e == null)
  return 0;

c.consume(e);
return 1 + drain(c, limit -1);
```

jctools-core/src/test/java/org/jctools/queues/QueueSanityTestMpscBlockingConsumerArrayExtended.java
```diff
@@ -375,11 +500,17 @@ public void testSpinWaitForUnblockForever() throws InterruptedException {
             }
 
             public void run() {
+                ArrayDeque<T> ints = new ArrayDeque<>();
                 try {
                     for (int i = 0; i < interations; ++i) {
                         T t;
                         do {
-                            t = source.poll(1,  TimeUnit.NANOSECONDS);
+                            if (useDrain) {
```
nitsanw, line 508: this test should be in the extended MPQ test IMO

nitsanw (changes requested):
see comments

jctools-core/src/test/java/org/jctools/queues/QueueSanityTestMpscBlockingConsumerArrayExtended.java
franz1981: I've moved it into the MPQ test but it's a shame will run twice because the queues needed by the test are allocated during the test itself and not upfront as the other MPQ tests

franz1981:
@nitsanw I've made the changes, thanks for the review, is much simpler now :)

jctools-core/src/test/java/org/jctools/queues/MpqSanityTestMpscBlockingConsumer.java
```diff
@@ -25,4 +30,59 @@ public MpqSanityTestMpscBlockingConsumer(ConcurrentQueueSpec spec, MessagePassin
         list.add(makeMpq(0, 1, SIZE, Ordering.FIFO, new MpscBlockingConsumerArrayQueue<>(SIZE)));// MPSC size SIZE
         return list;
     }
+
+    @Test(timeout = TEST_TIMEOUT)
+    public void testSpinWaitForUnblockDrainForever() throws InterruptedException {
```
nitsanw, line 35: why is this test not in the extended suite?

nitsanw: there's an MPQ extended test class now

nitsanw (approved):
LGTM, I'll tweak the test location and move on with release

PR #344, opened 2021-03-25
Fix table-size long math on very large tables.

cliffclick:
Fix reprobe rate for pathological keys.
Fix rare mis-count during table-promotion counting
Allow a fast not-atomic size-preserving clear, for rapid recycling of large tables.
Remove some IDE warnings.


nitsanw:
Thanks for the PR, I've been away this weekend but will have a look today :-)

nitsanw (approved):

nitsanw:
@cliffclick for future reference it is easier on reviewers if "whitespace" changes (e.g. renaming a variable, tidying up comments etc) are split into separate commits than functional changes (e.g. introducing a hash function, adding clone).

PR #346, opened 2021-04-02
Faster size on stable tables.

cliffclick:
Faster iteration on stable tables.


nitsanw:
Looks good, would be good to split this commit so that changes are more visible in the history

nitsanw:
@cliffclick I split the commits, it would be good to add test cases to cover the new code

PR #347, opened 2021-04-21
Update to JCStress 0.9

shipilev:
JCStress 0.8 and 0.9 comes with a few goodies that should be useful for concurrency testing: split compilation and actor affinity. This picks up 0.9.

Testing:
 - [x] `mvn clean install -pl jctools-concurrency-test`
 - [x] `java -jar jctools-concurrency-test/target/concurrency-test.jar` (`SingleWriterHashSetDuplicateReadsTest` still fails, as it should?).


nitsanw:
Thanks!

PR #366, opened 2022-10-20
safety checking

jessielesbian:


nitsanw:
Thanks for the PR.
As @franz1981 pointed out, this is a hot path, and the code expects callers to take precautions. Hence, Unsafe.

In addition, the code is incorrect in comparing ‘offset’ (in bytes from ref base) with ‘length’ (in elements).

nitsanw:
For future reference, please run tests before sending a PR.

PR #370, opened 2022-11-15
Fixes #367: generate a JPMS module descriptor

### LINKED ISSUE #367: Missing `Automatic-Module-Name` in 4.0.1 (regression)
Version 4.0.1 on Maven Central does not contain an `Automatic-Module-Name` (introduced in #289) in the manifest. I am not able to reproduce the problem compiling the `v4.0.1` tag from sources, but it must be caused by the `maven-jar-plugin` configuration on the host that built the released artifact.

Remark that the currently used `maven-bundle-plugin` version supports the automatic generation of JPMS module descriptors (even on JDK 8) through the `<_jpms-module-info/>` instruction. The generated module descriptor is functionally equivalent to the generated OSGI bundle descriptor. I can provide a PR that modularizes `jctools-core`, `jctools-channels` and `jctools-experimental`.

ppkarwasz:
Until now `jctools-core` used the Apache Felix Maven Bundle Plugin to generate its OSGI descriptor.

This PR:

* moves the internal classes to the `org.jctools.util.internal`. If needed this package can be exported to other JCTools artifacts using the `@ExportTo` annotation from BND.
* exports only explicitly annotated packages (i.e. all except `org.jctools.util.internal`). The OSGI descriptor becomes:
```
Export-Package: 
 org.jctools.queues.atomic;version="4.0.2";uses:="org.jctools.queues",
 org.jctools.queues.unpadded;version="4.0.2";uses:="org.jctools.queues",
 org.jctools.queues;version="4.0.2",
 org.jctools.maps;version="4.0.2",
 org.jctools.util;version="4.0.2",
 org.jctools.counters;version="4.0.2"
Import-Package: sun.misc;resolution:=optional
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
```
* automatically generates a `module-info.class` JPMS descriptor, which is in sync (cf. [JPMS libraries](https://bnd.bndtools.org/chapters/330-jpms.html)) with the OSGI descriptor:
```
module org.jctools.core@4.0.2.SNAPSHOT {
  requires java.base;
  requires static jdk.unsupported;
  exports org.jctools.queues;
  exports org.jctools.queues.atomic;
  exports org.jctools.queues.unpadded;
  exports org.jctools.maps;
  exports org.jctools.util;
  exports org.jctools.counters;
}
```
This descriptor can be generated by JDK 8.

### DIFF
--- a/jctools-core/pom.xml
+++ b/jctools-core/pom.xml
@@ -13,6 +13,12 @@
 	<packaging>bundle</packaging>
 
 	<dependencies>
+        <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>osgi.annotation</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
 		<dependency>
 			<groupId>org.hamcrest</groupId>
 			<artifactId>hamcrest-all</artifactId>
@@ -104,13 +110,6 @@
 						</goals>
 					</execution>
 				</executions>
-				<configuration>
-					<archive>
-						<manifestEntries>
-							<Automatic-Module-Name>org.jctools.core</Automatic-Module-Name>
-						</manifestEntries>
-					</archive>
-				</configuration>
 			</plugin>
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
--- a/jctools-core/src/main/java/org/jctools/counters/package-info.java
+++ b/jctools-core/src/main/java/org/jctools/counters/package-info.java
@@ -1,0 +1,17 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@Export
+package org.jctools.counters;
+
+import org.osgi.annotation.bundle.Export;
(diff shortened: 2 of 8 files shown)


jponge:
That'd be good, although I think in JPMS land you should not append a version but just use `org.jctools.core`

nitsanw:
@ppkarwasz I appreciate your effort here, but please minimize the contents of the PR to the absolute minimum.
Re-formatting of spaces and tabs, re-ordering of imports, changing unrelated pom files, all of these are not part of the value proposition of your PR. Please revise the PR to contain only the essential changes.

Thanks!

franz1981:
many thanks @nitsanw for the quick review!!

ppkarwasz:
Hi @nitsanw,

I refactored the PR to the bare minimum: the `maven-bundle-plugin` is reconfigured to export only packages explicitly marked as exported and to generate a `module-info.class` with a matching OSGI manifest.

> I'm confused about the value of moving the internal classes to `internal`, can you elaborate?

OSGI-wise it doesn't make sense: since the other artifacts in this repo use classes in `org.jctools.core`, you need to export the package to everybody.

The advantage appears on the JPMS side: using the `exports ... to` directive you can third-party modules from using the classes, while at the same time allowing your modules to use them. This is what JUnit 5 did with `junit-platform-common` (cf. [their `module-info.java` file](https://github.com/junit-team/junit5/blob/main/junit-platform-commons/src/module/org.junit.platform.commons/module-info.java)). Since `PaddedAtomicLong` must be exported for backward compatibility, it is easier to move the remaining classes.

ppkarwasz:
> That'd be good, although I think in JPMS land you should not append a version but just use `org.jctools.core`

@jponge: I just copied the result of `javap module-info.class` in the above comment. The compiled module descriptor has a version attribute, but I don't think it can be used for anything practical.

jponge:
I've checked this PR, the generated `module-info` looks good:

```
module org.jctools.core {
    requires java.base;
    requires jdk.unsupported;

    exports org.jctools.maps;
    exports org.jctools.util;
    exports org.jctools.queues;
    exports org.jctools.queues.atomic;
    exports org.jctools.queues.unpadded;
    exports org.jctools.counters;
}
```

franz1981:
@nitsanw do you think it's ok to roll a new release and get this in? 
AFAIK @jponge as others could rely on the presence to dep on JCTools

franz1981:
@nitsanw If I can help you somehow, just ping me via email (I've sent you an email some time ago, so you'll likely see that one too ;) ) that I will happily help

Issue #355, opened 2021-12-28
OOM can cause linked array qs to break


franz1981:
Not high priority given that I won't suggest users to run *any* application after receiving an OOM, but still relevant: [here](
https://github.com/netty/netty/issues/11956#issuecomment-1002041097) an explanation on how it could happen.

It could be *easily* solved by changing semantic of poll, peek and isEmpty during a resizing:
On an empty q, If the parity bit is just set and the new linked chunk is not yet set, isEmpty should return true and both poll and peek should return null.
This allow a producer index rollback to pass unnoticed because resizing won't make a consumer to believe that there's some data.
Data will be perceived as available just when producer index is further moved (and by consequence, a JUMP ready to be spin-awaited)

TLDR any emptiness check should use both consumer and producer index >> 1 to save considering any parity bit due to resizing

@nitsanw other ideas?

nitsanw:
I think we can fix the spin loop to read the updated `pIndex` in the loop so it can recover, or we can take the current `size` policy and extend it. 

_\<rambling\> I'm a little uncomfortable with the recovery logic violating the monotonically increasing property of the indexes. It's a nice property to have, but I can also see the argument that the parity bit is not REALLY an advance in this usecase, we could have used the sign bit for the same purpose and that would have violated monotonic progress just as much.\</rambling\>_

Proposed solution, we follow the `size` current behaviour here:
```
size = (cIndex - pIndex)/2 -> queue in transition is empty

=> isEmpty = (cIndex - pIndex)/2 == 0 (make it match size, but without all the funky mitigations)

=> poll/peek loop needs to change:
            long pIndex = lvProducerIndex();
            if (index - pIndex)/2 == 0) return null;
            // poll() == null iff queue is empty, null element is not strong enough indicator, so we must
            // check the producer index. If the queue is indeed not empty we spin until element is
            // visible.
            do
            {
                e = lvRefElement(buffer, offset);
            }
            while (e == null);
```

franz1981:
Agree: I like the proposed solution and I was thinking about re-reading pIndex every X spin for element as well (very happy we reached the same exact conclusions), although I believe that adding a chunk could be a slow operation (if the allocation is outside a tlab, for example) and being able to return back to the user application code (with a null poll) can help similarly to relaxedPoll to give the application code the chance to do something different (if any)

PR #371, opened 2023-02-08
Update BaseMpscLinkedArrayQueue.java

huyang327:
Let the poll method obtain data as soon as possible, and at the same time prevent the OOM problem that occurs after the resize method is started during the offer process, resulting in an infinite loop of the poll method.

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/BaseMpscLinkedArrayQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/BaseMpscLinkedArrayQueue.java
@@ -296,21 +296,14 @@ abstract class BaseMpscLinkedArrayQueue<E> extends BaseMpscLinkedArrayQueueColdP
 
         final long offset = modifiedCalcCircularRefElementOffset(cIndex, mask);
         Object e = lvRefElement(buffer, offset);
-        if (e == null)
+        while (e == null)
         {
-            long pIndex = lvProducerIndex();
-            // isEmpty?
-            if ((cIndex - pIndex) / 2 == 0)
-            {
-                return null;
-            }
-            // poll() == null iff queue is empty, null element is not strong enough indicator, so we must
-            // spin until element is visible.
-            do
+            if (cIndex != lvProducerIndex())
             {
                 e = lvRefElement(buffer, offset);
+                continue;
             }
-            while (e == null);
+            return null;
         }
 
         if (e == JUMP)


franz1981:
This is related https://github.com/netty/netty/issues/13137 right @huyang327 ?

jctools-core/src/main/java/org/jctools/queues/BaseMpscLinkedArrayQueue.java
```diff
@@ -296,21 +296,14 @@ public E poll()
...
-            long pIndex = lvProducerIndex();
-            // isEmpty?
-            if ((cIndex - pIndex) / 2 == 0)
-            {
-                return null;
-            }
-            // poll() == null iff queue is empty, null element is not strong enough indicator, so we must
-            // spin until element is visible.
-            do
+            if (cIndex != lvProducerIndex())
```
nitsanw, line 301: This is reverting the fix for size/isEmpty compatibility

franz1981:
@huyang327 

So you gather new info about that case? you see OOM there? If Yes, please update the case.
This fix is breaking the new assumption that a mid-rotation (that set a parity bit) means that there's no value i.e. emptyness, that's has been enforced due to https://github.com/JCTools/JCTools/issues/355

nitsanw:
@huyang327 Thanks for the PR, but I'm reluctant to accept because:
* It's not clear which problem is being solved if any. Can you provide an issue, a test or a step by step demonstration that there is an issue? How would it manifest to users?
* The commit message is very vague.

So at this point I don't really understand what this is supposed to solve...

huyang327:
> @huyang327 Thanks for the PR, but I'm reluctant to accept because:
> 
> * It's not clear which problem is being solved if any. Can you provide an issue, a test or a step by step demonstration that there is an issue? How would it manifest to users?
> * The commit message is very vague.
> 
> So at this point I don't really understand what this is supposed to solve...

If the OOM problem occurs when the resize method is not considered, the modification I submitted is really meaningless. I see that the cjtools4.0.1 version has modified this problem, and the modification plan is to feed back null in advance.
Object e = lvRefElement(buffer, offset);
if (e == null)
         {
             long pIndex = lvProducerIndex();
             // isEmpty?
             if ((cIndex - pIndex) / 2 == 0)
             {
                 return null; (<============================ here)
             }
             // poll() == null iff queue is empty, null element is not strong enough indicator, so we must
             // spin until element is visible.
             do
             {
                 e = lvRefElement(buffer, offset);
             }
             while (e == null);
         }
My idea is to try to get the data when the poll method is called when possible. Because if ((cIndex - pIndex) / 2 == 0), the queue may be running the offer method to add data, for example, the resize method is being called to add JUMP. I want to try to read the data as fast as possible through the loop.

franz1981:
@huyang327 

> the queue may be running the offer method to add data, for example, the resize method is being called to add JUMP.

That's the point: if a new chunk is going to be allocated, there's no data yet to be consumed and spinning would just consume CPU (to not mention the OOM issue we've talked about); we've decided to "fail-fast" instead exactly to give the user the choice what to do: effectively an offer that perform an allocation is NOT 100% sure will complete.

huyang327:
Understand, the principle is to fail fast and I have no doubts. Thanks
@franz1981 @nitsanw

PR #388, opened 2024-02-06
Adding UnpaddedAtomic queues

### LINKED ISSUE #387: Unpadded variants could avoid using Unsafe
HI @nitsanw !

Now, for convenience we create the unpadded version of the queues out of the Unsafe-based ones, but actually, if users doesn't care about maximum performance by preventing false-sharing, maybe they are NOT interested either to use Unsafe as well, for the same reason. This would make the adoption of such wider, and life of native-image users, happier, wdyt?

franz1981:
Fixes #387 

@nitsanw I didn't tried yet to reuse some unpadded gen code, but the most of the work should be done, if the naming is fine for you.


franz1981:
@jponge ;)

franz1981:
I see now that I have created the queues under unpadded.atomic instead of atomic.unpadded, likely because I've decided for names like UnpaddedAtomicArrayQueue, because I love to keep AtomicQueue together, but if you prefer differently, I can do it; i really have no strong preferences on this

franz1981:
@nitsanw If you prefer `atomic.unpadded` I created an alternative branch (and can squash the commits if is better) ->  https://github.com/franz1981/JCTools/tree/atomic_unpadded

nitsanw:
There's already an unpadded code generator, I think it would be better to enhance it rather than add code to all the different atomic queue generators.
Also, please stick with `atomic.unpadded`.
Thanks!

franz1981:
> There's already an unpadded code generator,

In term of code changes, and complexity, I found much easier to modify the atomic generators, given that the most of the work is there, while unpadding seems a relatively easier job (mostly removing comments and unpadded fields).
I will open a PR based on https://github.com/franz1981/JCTools/tree/atomic_unpadded

PR #389, opened 2024-02-07
Adding Atomic Unpadded queues

franz1981:
This is replacing https://github.com/JCTools/JCTools/pull/388 

Right now it modifies the atomic generators, which are doing already the heavy-lifting of replacing methods/primitive utils and classes to allow `Atomic`s to be used, but adding the 2 required behaviour:
1. removing padding comments
2. removing padding fields

It looked like a natural thing to me, given that it would add unpadding to the existing atomic generation mechanism.

As promised on https://github.com/JCTools/JCTools/pull/388#issuecomment-1931845891, I'm exploring how to modify the existing `JavaParsingUnpaddedQueueGenerator` to create `Atomic` variants, but TBH it seems to add more changes (probably taking them - by decoration or copy - from the existing Atomic generators)


franz1981:
IDK if e1d445eb0e1a312eae62a0d354af4d9e0809cf85 vs this one matches some of your requests, because i didn't enhanced the unpadded generator, because the most of the parts I needed to reuse was from the atomic generators, while unpadding just requires to remove unpadded fields.
I'm not a big fan of inheritance, in general, but I've replaced the changes in the atomic generator with some extension points to allow the atomic unpadded generators to inject the package/queue names they needed, in order to reuse as much as possible what we already have.

I tried to extract the parts from the atomic generator, to enhance the unpadded one, without much luck, due to lack of time (admit), but if you have any hint I'm open to do things differently, bud

jctools-build/src/main/java/org/jctools/queues/util/GeneratorUtils.java
```diff
@@ -0,0 +1,94 @@
...
+        for (int i = 1; i < args.length; i++) {
+            File file = new File(args[i]);
+            System.out.println("Processing " + file);
+            CompilationUnit cu = new JavaParser().parse(file).getResult().get();
+            JCToolsGenerator generator = buildGenerator(generatorClass, file.getName());
+            generator.visit(cu, null);
+            generator.organiseImports(cu);
+            generator.cleanupComments(cu);
+
+            String outputFileName = generator.translateQueueName(file.getName().replace(".java", "")) + ".java";
```
franz1981, line 85: I could make `JCToolsGenerator` to use a single method `String generate(String queueName)` which return the translated queue name instead of having 3 different methods, executing a generic generation process

franz1981: I'm saving to add more abstractions @nitsanw because I don't know yet if this PR is decent enough; IDK if makes sense to create abstractions of any kind, if not for the sake of re-using code, given that generating code isn't a proper domain which require being strongly defined, but I see it more as a serie of transformation steps/scripts which we want to write once and modify after 4-5 years (unless bugs).

jctools-core/src/main/java/org/jctools/queues/atomic/unpadded/AtomicReferenceArrayQueue.java
```diff
@@ -0,0 +1,158 @@
...
+
+import java.util.AbstractQueue;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+
+import static org.jctools.queues.atomic.unpadded.AtomicQueueUtil.calcCircularRefElementOffset;
+import static org.jctools.queues.atomic.unpadded.AtomicQueueUtil.lvRefElement;
+
+abstract class AtomicReferenceArrayQueue<E> extends AbstractQueue<E> implements IndexedQueue, QueueProgressIndicators, MessagePassingQueue<E>, SupportsIterator
```
nitsanw, line 31: We can avoid the copy by making the original public and annotating as `InternalAPI`

jctools-core/src/main/java/org/jctools/queues/atomic/unpadded/LinkedQueueAtomicNode.java
```diff
@@ -0,0 +1,69 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jctools.queues.atomic.unpadded;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+public final class LinkedQueueAtomicNode<E> extends AtomicReference<LinkedQueueAtomicNode<E>>
```
nitsanw, line 18: Can reuse original which is unpadded

nitsanw (changes requested):
Overall I think it looks OK, please avoid copying stuff where possible, even if it requires changing class visibility (just mark as internal)

franz1981:
Nice one, ok @nitsanw I will reuse as much as we can :+1:  thanks for taking a look!

franz1981:
@nitsanw PTAL

jctools-core/pom.xml
```diff
@@ -173,6 +173,57 @@
                             </arguments>
                         </configuration>
                     </execution>
+					<execution>
```
nitsanw, line 176: What's happening with the indentation here?

Jonathanvara (approved):

PR #309, opened 2020-06-01
Fixes #308: Remove useless fences on chunk's prev field accesses

franz1981:

PR #394, opened 2024-11-30
Fix MpUnboundedXaddChunk.soPrev

chrisvest:
The soPrev method implies that it's making a store with RELEASE semantics, but it was doing a plain store.

This changes the method implementation to store with RELEASE, matching its naming convention. Since is also a more correct counterpart to the only loads of this field, which are always done with volatile memory semantics.

It's not clear to me if this fixes a real issue on architectures with weaker memory orderings than x86.

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpUnboundedXaddChunk.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpUnboundedXaddChunk.java
@@ -67,7 +67,7 @@ public class MpUnboundedXaddChunk<R,E>
 
     public final void soPrev(R value)
     {
-        UNSAFE.putObject(this, PREV_OFFSET, value);
+        UNSAFE.putOrderedObject(this, PREV_OFFSET, value);
     }
 
     public final void soElement(int index, E e)


chrisvest:
FYI @franz1981

franz1981:
Thanks @chrisvest for looking into this, so...
According to https://github.com/JCTools/JCTools/blob/a17b56f09cc9ec7d773707578c89f3ca9c11f968/jctools-core/src/main/java/org/jctools/queues/MpUnboundedXaddArrayQueue.java#L370 and https://github.com/JCTools/JCTools/blob/a17b56f09cc9ec7d773707578c89f3ca9c11f968/jctools-core/src/main/java/org/jctools/queues/MpUnboundedXaddArrayQueue.java#L389

In both cases there are other operations which release the prev value (along with other states) hence it could be that the prev store could be a plain one too: this means that the method name was wrong, likely.
I have re-read my own impl to make sure of this, but I see that 
https://github.com/JCTools/JCTools/blob/a17b56f09cc9ec7d773707578c89f3ca9c11f968/jctools-core/src/main/java/org/jctools/queues/MpUnboundedXaddArrayQueue.java#L323 too is not required to be a load acquire but just a plain one, since the queue logic assume that the prev values are released by https://github.com/JCTools/JCTools/blob/a17b56f09cc9ec7d773707578c89f3ca9c11f968/jctools-core/src/main/java/org/jctools/queues/MpUnboundedXaddArrayQueue.java#L350

franz1981:
And indeed I already found this but forgot to merge it, see https://github.com/JCTools/JCTools/pull/309

chrisvest:
Five years ago, lol. Yeah, maybe #309 need to be brought over the finish line.

PR #400, opened 2025-06-02
fix: make producer index immedately visible to consumer

leisurelyrcxf:
The contract is that if offer succeeds, then poll must see the offered object. Using soProducerIndex breaks such contract.

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MessagePassingQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MessagePassingQueue.java
@@ -107,7 +107,7 @@ public interface MessagePassingQueue<T>
      * according to the {@link Queue#offer(Object)} interface.
      *
      * @param e not {@code null}, will throw NPE if it is
-     * @return true if element was inserted into the queue, false iff full
+     * @return true if element was inserted into the queue, false if full or seen as full.
      */
     boolean offer(T e);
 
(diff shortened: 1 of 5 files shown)


jctools-core/src/main/java/org/jctools/queues/MessagePassingQueue.java
```diff
@@ -107,7 +107,7 @@ interface ExitCondition
      * according to the {@link Queue#offer(Object)} interface.
      *
      * @param e not {@code null}, will throw NPE if it is
-     * @return true if element was inserted into the queue, false iff full
+     * @return true if element was inserted into the queue, false if full or seen as full.
```
leisurelyrcxf, line 110: Let's see the case of mpsc queue, the consumer uses `soConsumerIndex`, which doesn't guarantee producer sees the latest consumption, thus producer may return false after consumption.

franz1981:
> If program depends on sequential consistency between poll/offer

These queues doesn't offer sequential consistent behavior but just happens-before/synchronized-with.

In short, no need to make producer index to be volatile.

leisurelyrcxf:
@franz1981 given resize() is such a low frequency operation, i think this is worthy to provide stronger memory semantic.

leisurelyrcxf:
@franz1981 @nitsanw 

closed but I think it is clearer if you made clearer statements in the doc

```
    /**
     * Called from the consumer thread subject to the restrictions appropriate to the implementation and
     * according to the {@link Queue#poll()} interface.
     *
     * @return a message from the queue if one is available, {@code null} iff empty
     */
    T poll();
```

by stating that the poll method doesn't guarantee a SC semantic, that is poll is not guaranteed to see a recent offer() in a SC way.

PR #413, opened 2026-03-20
Introduce MpscBlockingQueue

nicholas-keene-striim:
MpscBlockingQueue written by Nicholas Keene of Striim based on JCTools MpscCompoundQueue.

This queue efficiently receives messages on multiple threads and blocks when a queue size is surpassed. The size calculation is not exact, as it cannot be and our needs do not require it. Our needs were to block writers 
when the size grew large and to signal when the queue became empty.

The second of two commits in this pull request segregate the awaitEmpty() method which is built on a ConcurrentBoolean class. If JCTools doesn't want that method it can do without that additional class and without the second commit.

### DIFF
--- a/jctools-core/src/main/java/org/jctools/queues/MpscBlockingQueue.java
+++ b/jctools-core/src/main/java/org/jctools/queues/MpscBlockingQueue.java
@@ -1,0 +1,389 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jctools.queues;
+
+import static org.jctools.util.PortableJvmInfo.CPUs;
+import static org.jctools.util.Pow2.isPowerOfTwo;
+import static org.jctools.util.Pow2.roundToPowerOfTwo;
+
+import org.jctools.queues.MpscArrayQueue;
+import org.jctools.util.RangeUtil;
+
+import java.util.AbstractQueue;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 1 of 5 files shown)


nitsanw:
Hi! Can you please get the code to the project standard? tests, benchmarking, naming etc.

nicholas-keene-striim:
I added snapshot test, sanity test, and I played around with the benchmarks for a long time and left some tests I found useful. The throughput seemed _highly_ dependent on test parameters, showing test degradation in scenarios that my company has found to be optimal (large queues). As always your mileage may vary. If there is a doc explaining submission standards I didn't find it, but any feedback on the naming is welcome. Could someone from the project comment whether or not this queue paradigm is a good fit in the library?

jctools-core/src/main/java/org/jctools/queues/MpscBlockingQueue.java
```diff
@@ -0,0 +1,402 @@
...
+
+/**
+ * A blocking MPSC queue implementation based on JCTools MpscArrayQueue.
+ * This queue uses multiple parallel queues to reduce contention among producers.
+ *
+ * @param <E> the type of elements held in this queue
+ */
+public class MpscBlockingQueue<E> extends MpscBlockingQueueL2Pad<E> implements BlockingQueue<E>
+{
+    private ConcurrentBoolean isEmpty = new ConcurrentBoolean(true);
```
franz1981, line 150: why is not final?

jctools-core/src/main/java/org/jctools/queues/MpscBlockingQueue.java
```diff
@@ -0,0 +1,402 @@
...
+    private final AtomicInteger threadQueueIndexCounter = new AtomicInteger(0);
+    private final ThreadLocal<Integer> threadQueueIndex = new ThreadLocal<Integer>()
```
franz1981, line 152: this won't work well with virtual threads sadly :"/ (perf wise)

nitsanw:
> If there is a doc explaining submission standards I didn't find it, but any feedback on the naming is welcome. Could someone from the project comment whether or not this queue paradigm is a good fit in the library?

Sadly we have no standards documents, but you can follow the example of pre-existing queues:
- The queue name should be `MpscCompoundBlockingQueue` (or similar)
- It should probably implement the MPQ interface and have a matching test
- Javadoc should clarify similarities/differences from MpscCompoundQueue

If you ran benchmarks, please share the data. Compare with MPSC variants etc.

Thanks :)

nitsanw:
Please rebase as code style and testing github actions have been added as gates

nicholas-keene-striim:
We leave this code as an example of a useful thing that industry has done with the excellent JCTools codebase. Thank you.

PR #415, opened 2026-04-18
fix: JPMS module names incorrectly derived from OSGI bundle

### LINKED ISSUE #414: Module name is invalid for jctools-core-jdk11
The module name `org.jctools.core-jdk11` is not conformant to JPMS. The module name cannot have hyphens, you should consider shipping with a module-info to resolve this.

kermandev:
Fixes #414 

Used https://bnd.bndtools.org/instructions/jpms_module_info_options.html, and https://bnd.bndtools.org/macros/replace.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,8 +190,10 @@
                 </dependencies>
                 <configuration>
                     <instructions>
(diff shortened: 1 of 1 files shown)


nitsanw (approved):
LGTM

PR #416, opened 2026-04-23
Generate atomic/unpadded/varhandle variants for xadd queue family

franz1981:
Extend the existing code generators to produce all 5 variant types (atomic, atomic-unpadded, unpadded, varhandle, varhandle-unpadded) for MpUnboundedXaddArrayQueue, MpUnboundedXaddChunk, and their Mpsc/Mpmc specializations. This was the only queue family without generated variants.

Key generator changes:
- Handle Chunk class name translation in translateQueueName()
- Support getAndAdd/getAndIncrement method accessor patterns
- Add xadd-specific field names (producerChunk, consumerChunk, etc.)
- Skip final fields during accessor method patching
- Preserve type arguments when replacing SpscArrayQueue references
- Guard against double-translation of already-translated names
- Translate static imports from Chunk classes

Source file adjustments:
- Move $gen:ordered-fields directive on MpUnboundedXaddChunk
- Remove SUPPORTS_GET_AND_ADD_LONG conditional in base queue
- Replace Arrays.fill with soLongElement loop in MpmcUnboundedXaddChunk
- Add calcLongElementOffset to VarHandleQueueUtil


jctools-core-jdk11/src/main/java/org/jctools/queues/varhandle/VarHandleQueueUtil.java
```diff
@@ -87,6 +87,10 @@ public static long lvLongElement(long[] buffer, long offset) {
     return (long) ARRAY_LONG_ELEMENT.getVolatile(buffer, (int)offset);
   }
 
+  public static long calcLongElementOffset(long index) {
```
franz1981, line 90: This really sucks :/

jctools-build/src/main/java/org/jctools/queues/atomic/JavaParsingAtomicArrayQueueGenerator.java
nitsanw, line 1: Please add some comments for clarification

jctools-build/src/main/java/org/jctools/queues/atomic/JavaParsingAtomicQueueGenerator.java
nitsanw, line 1: Please add some comments to new code

nitsanw (commented):
Please document the parser changes, I will review the generated code just now

nitsanw:
Where are the tests?

franz1981:
> Where are the tests?

I haven't yet updated the existing tests to cover for the new impls, good point @nitsanw

franz1981:
@nitsanw addressed your comments and run the tests on the new variants - just one note: I have noticed we have no github workflow to handle tests on PRs as well (running on ARM, x86 and MacOS) - do you want to me add it to JCTools? It seems a good idea

nitsanw:
Sound like a good idea.

franz1981:
> Sound like a good idea.

I would do it as a follow up or in parallel, if you agree!
Let me know if i'm missing something for this PR or is fine

jctools-core-jdk11/src/main/java/org/jctools/queues/varhandle/unpadded/MpUnboundedXaddVarHandleUnpaddedArrayQueue.java
```diff
@@ -0,0 +1,387 @@
...
+
+    private volatile long producerIndex;
+
+    @Override
+    public final long lvProducerIndex() {
+        return producerIndex;
+    }
+
+    final long getAndIncrementProducerIndex() {
+        return (long) VH_PRODUCER_INDEX.getAndAdd(this, 1);
```
franz1981: this_ is not good! It should use 1L :"(

VH polymorphic sig is TERRIBLE @nitsanw I have to fix this

franz1981:
PTAL again @nitsanw

jctools-build/src/main/java/org/jctools/queues/atomic/JavaParsingAtomicArrayQueueGenerator.java
```diff
@@ -93,11 +94,40 @@ String fieldUpdaterFieldName(String fieldName) {
...
+        case "next":
+            return "NEXT_UPDATER";
         default:
             throw new IllegalArgumentException("Unhandled field: " + fieldName);
         }
     }
 
+    /**
+     * Replaces {@code new SpscArrayQueue<R>(...)} with the unpadded pool queue variant appropriate
+     * for this generator (e.g. {@code new SpscAtomicUnpaddedArrayQueue<R>(...)}).
```
nitsanw, line 117: Why unpadded? The unsafe variant is padded

franz1981: I changed that one (I have to check it applies to the existing unsafe variant) because is not supposed to be an hot path and have a bigger footprint

nitsanw (approved):
OK, I guess

PR #424, opened 2026-06-22
Add CI test workflow for x86_64 Linux, ARM64 Linux, and ARM64 macOS

franz1981:
Run mvn test on jctools-build, jctools-core, and jctools-core-jdk11 across ubuntu-24.04 (x86_64), ubuntu-24.04-arm (aarch64), and macos-15 (Apple Silicon) with JDK 11 and 21.

ARM runners are free for public repos and exercise weak memory ordering that x86/TSO masks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: CI
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  test:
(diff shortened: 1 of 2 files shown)


He-Pin:
true， Cat Effect was there a bug only on ARM boxes

franz1981:
@nitsanw shame on me I've wrongly pushed https://github.com/JCTools/JCTools/commit/5c13136cbff1fb701fc7241291151cb514bf2086 into master :cry:

PR #425, opened 2026-07-06
Reject queueParallelism < 1 in MpscCompoundQueue

raphaelroshan:
`new MpscCompoundQueue(capacity, 0)` constructed a corrupt queue: `isPowerOfTwo(0)` returns `true`, so `parallelQueues` became `0`, `parallelQueuesMask` became `-1`, and `queues` was a zero-length array. Construction succeeded, but the first `offer()` indexed the empty array and threw `ArrayIndexOutOfBoundsException`.

Validates `queueParallelism >= 1` up front with `RangeUtil`, mirroring the existing `fullCapacity` check, so a misconfiguration fails fast at construction. Adds a test for the rejection and a minimal-parallelism round-trip.


jctools-core/src/test/java/org/jctools/queues/MpscCompoundQueueTest.java
```diff
@@ -0,0 +1,27 @@
...
+    @Test
+    public void offerPollRoundTripsWithMinimalParallelism()
+    {
+        MpscCompoundQueue<Integer> queue = new MpscCompoundQueue<Integer>(1024, 1);
+        assertTrue(queue.offer(1));
+        assertTrue(queue.offer(2));
+        assertEquals(Integer.valueOf(1), queue.poll());
+        assertEquals(Integer.valueOf(2), queue.poll());
+        assertNull(queue.poll());
+    }
```
nitsanw: This is redundant if covered(and should be) by the QueueSanityTest

nitsanw (commented):
LGTM overall, please check my comment on the test

raphaelroshan:
Thanks for the feedback, updated! The offer/poll round-trip is already covered by QueueSanityTestMpscCompound, so the test class now just keeps the zero-parallelism guard.

nitsanw (approved):
LGTM

nitsanw:
Thanks @raphaelroshan :)

raphaelroshan:
Thank you! @nitsanw

